PR #2, opened 2022-10-31
cli: add print formats, inform user that files are written

boeddeker:
Open points:
 - Should, by default, the file `details.json` and `average.json` be written to `cwd`?
 - Should it be possible to print details to stdout? e.g. `--details`
 - `--json` and `--yaml` could be replaced with something like `--format [json|yaml|txt]`. I am not sure, which I like more.

 - Another thing, should we provide a CLI tool for WER operations? e.g.
   - `details_subset1.json` plus `details_subset2.json`
   - `details.json` to `average.json`

 - Is there an advantage, to calculate cpwer, orcwer and mimower with a single call, except the loading of the files? It might be better for future extensions to use commands (e.g. `python -m meeteval.wer orc`) instead of flags.

### DIFF
--- a/meeteval/wer/__main__.py
+++ b/meeteval/wer/__main__.py
@@ -1,4 +1,5 @@
 """Computes the Concatenated minimum Permutation Word Error Rate (cpWER)"""
+import sys
 import argparse
 import dataclasses
 import json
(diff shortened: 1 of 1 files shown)


boeddeker:
We discussed this offline and did a mayor refactoring of the cli.

PR #8, opened 2023-03-08
Use argparse for CLI

thequilo:


meeteval/wer/__main__.py
```diff
@@ -58,12 +58,12 @@ def _load(path: Path):
             raise NotImplemented(f'Unknown file ext: {path.suffix}')
 
 
-def _load_reference(reference: Path):
+def _load_reference(reference: [Path, str, io.TextIOBase, tuple, list]):
```
boeddeker, line 61: Change to `'Path | str | io.TextIOBase | tuple | list'`. I think that is easier to read and will be the future style (The quotes allow it to be used in python version below 3.10).

meeteval/wer/__main__.py
```diff
@@ -381,32 +349,46 @@ def load(file: Path):
...
+    # Parse arguments and find command to execute
+    args = parser.parse_args()
+
+    if hasattr(args, 'func'):
+        kwargs = vars(args)
+        fn = args.func
+        # Pop also removes from args namespace
+        kwargs.pop('func')
+        kwargs.pop('version')
+        fn(**kwargs)
```
boeddeker, line 383: Change to `return fn(**kwargs)`.

meeteval/wer/__main__.py
```diff
@@ -189,174 +181,150 @@ def _save_results(
...
-                 'of the (first) hypothesis file and {wer} is replaced with the wer'
-                 'type (equal to the command name). '
 
+def command(name, help, parent_parsers):
+    def _command(fn):
+        command_parser = commands.add_parser(
+            name,
+            add_help=False,
+            formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+            help=help,
```
boeddeker, line 242: Can you change this to `fn.__doc__`?

meeteval/wer/__main__.py
```diff
@@ -189,174 +181,150 @@ def _save_results(
...
+)
+common_wer_parser.add_argument(
+    '-h', '--hypothesis', help='Hypothesis file in STM format',
+    # type=argparse.FileType('r'),
+    nargs='+',
+    action='append',
+)
+common_wer_parser.add_argument(
+    '--average-out',
+    default='{parent}/{stem}_{wer}.json',
```
boeddeker, line 210: Would the default be set with `command_parser.set_defaults(func=fn)`, when the function defines a default? If yes, could you change it, that the function defines the default?

boeddeker (approved):

PR #11, opened 2023-03-24
Add time-constrained minimum-permutation WER

thequilo:
This PR adds a time-constrained version of the cpWER which could be useful for evaluation of Chime-7. 

It performs matching using the words and word- or segment-level time annotations. By doing this, we can make sure that words are not matched between utterances or segments that are far apart temporally.

It can be used with 

```bash
# Example with collar of 5s for the hypothesis
meeteval-wer tcpwer -h hyp.stm -r ref.stm --hyp-collar 5
```


doc/tcpwer.md
```diff
@@ -0,0 +1,37 @@
...
+- It should not be penalized when the system combines several words (e.g. an utterance) in one segment, but
+- The system should be penalized when it produces long segments spanning multiple reference segments
+- It should not be penalized when the system provides more precise timing than the reference (e.g., by splitting in a pause or producing tighter bounds)
+
+## Pseudo-word-level annotations
+To compute the matching, we need a temporal annotation (start and end time) for each word.
+Often, detailed word-level temporal annotations are not available, either because annotating a reference is expensive or a system does not produce such detailed information.
+We thus implement different strategies to infer "pseudo-word-level" timings from segment-level timings:
+
+- `full_segment`: Copy the time annotation from the segment to every word within that segment
```
boeddeker, line 20: Mention, that this is good for ref and bad for hyp?

ref: Humans might group words with long pauses in between, e.g. in LibriSpeech, where ...

hyp: `full_segment` can be fooled (e.g. set start to 0 and end to a large number to get the cpWER), hence avoid it.

Have we already a suggestion, if `equidistant_intervals` or `euqidistant_points` should be used?

doc/tcpwer.md
```diff
@@ -0,0 +1,37 @@
...
+Due to the way we estimate pseudo-word-level annotations for the segment-level annotations, the collar has to be relatively large (compared to typical values for DER computation).
+It should be chosen so that small diarization errors (e.g., merging two utterances of the same speaker uttered without a pause into a single segment) are not penalized but larger errors (merging utterances that are tens of seconds apart) is penalized.
+This, of course, depends on the data, but we found values in the range of 2-5s to work well in some test on libri-CSS.
+
+## Using tcpWER
+
+The tcpWER currently only support STM files because they provide all necessary information on a segment level.
+You can use any resolution for the begin and end times (e.g., seconds or samples), but make sure to adjust the collar accordingly (`5` or `80000` for 16kHz).
+```shell
+meeteval-wer tcpwer -h hyp.stm -r ref.stm --hyp-collar 5
```
boeddeker, line 36: What is the default `Pseudo-word-level annotations`?

meeteval/io/stm.py
```diff
@@ -72,6 +72,16 @@ def serialize(self):
         return (f'{self.filename} {self.channel} {self.speaker_id} '
                 f'{self.begin_time} {self.end_time} {self.transcript}')
 
+    def segment_dict(self):
+        """Returns a segment dict"""
```
boeddeker, line 76: I changed my mind, I think we should mention, that this is CHiME-7 style, although the type of start_time and end_time is wrong.
Otherwise, the field names are strange. `session_id` is CHiME specific, and it will likely confuse a user, that we changed most key names.

meeteval/wer/matching/cy_levenshtein.pyx
```diff
@@ -0,0 +1,279 @@
...
+            uint cost_del,
+            uint cost_ins,
+            uint cost_sub,
+            uint cost_cor,
+            uint eps
+    )
+
+
+def obj2vec(a, b):
+    int2sym = dict(enumerate(sorted(set(a) | set(b))))
```
boeddeker, line 69: Add https://github.com/pzelasko/kaldialign/blob/17d2b228ec575aa4f45ff2a191fb4716e83db01e/kaldialign/__init__.py#L6-L15 as reference for the original code

meeteval/wer/matching/cy_levenshtein.pyx
```diff
@@ -0,0 +1,279 @@
...
+    return levenshtein_distance_(reference, hypothesis)
+
+def levenshtein_distance_custom_cost(
+        reference,
+        hypothesis,
+        cost_del,
+        cost_ins,
+        cost_sub,
+        cost_cor,
+):
```
boeddeker, line 88: How about merging `levenshtein_distance` and `levenshtein_distance_custom_cost` and add a comment in the code, that `levenshtein_distance_` is faster?

meeteval/wer/matching/cy_levenshtein.pyx
```diff
@@ -0,0 +1,279 @@
...
+    assert len(reference) == 0 or reference_timing.shape == (len(reference), 2), (reference_timing.shape, len(reference))
+    assert len(hypothesis) == 0 or hypothesis_timing.shape == (len(hypothesis), 2), (hypothesis_timing.shape, len(hypothesis))
+    assert len(reference) == 0 or len(hypothesis) == 0 or reference_timing.dtype == hypothesis_timing.dtype, (reference_timing.dtype, hypothesis_timing.dtype)
+
+    def check(timing):
+        # end >= start
+        if np.any(timing[:, 1] < timing[:, 0]):
+            raise ValueError(f'The end time of an interval must not be smaller than its begin time')
+        # start values are increasing
+        if np.any(np.diff(timing[:, 0]) < 0):
```
boeddeker, line 119: Does the code work correctly, when the "next" end_time is smaller?

meeteval/wer/matching/cy_levenshtein.pyx
```diff
@@ -0,0 +1,279 @@
...
+            raise ValueError(
+                f'The start times of the annotations must be increasing, which they are not. '
+                f'This might be caused by overlapping segments, see the (potential) previous warning.'
+            )
+
+    if len(reference):
+        check(reference_timing)
```
boeddeker, line 126: Has cython a informative traceback? Otherwise, the information, whether ref or hyp is buggy is lost.

meeteval/wer/matching/cy_levenshtein.pyx
```diff
@@ -0,0 +1,279 @@
...
+            cost_del,
+            cost_ins,
+            cost_sub,
+            cost_cor,)
+    if np.issubdtype(reference_timing.dtype, np.signedinteger):
+        return time_constrained_levenshtein_distance_[int](*args)
+    elif np.issubdtype(reference_timing.dtype, np.unsignedinteger):
+        return time_constrained_levenshtein_distance_[uint](*args)
+    elif np.issubdtype(reference_timing.dtype, np.floating):
+        return time_constrained_levenshtein_distance_[float](*args)
```
boeddeker, line 161: Is this the python float (double precision floating point number with 64 bits) or the c float (single precision floating point number with 32 bits)?

boeddeker: We should use double precision

tests/test_cli.py
```diff
@@ -76,6 +76,12 @@ def test_burn_cp():
     run(f"python -m meeteval.wer cpwer -h 'hyp?.stm' -r 'ref?.stm'")
 
 
+def test_burn_tcp():
+    run(f'python -m meeteval.wer tcpwer -h hyp.stm -r ref.stm')
+    run(f'python -m meeteval.wer tcpwer -h hyp.stm -r ref.stm --hyp-collar 1 --ref-collar 1')
```
boeddeker, line 81: Use `--hyp-collar 5 --ref-collar 0`? Such that it matches our recommendation.

meeteval/wer/matching/cy_levenshtein.pyx
thequilo: Yes, I now keep track of the largest seen end value for the pruning optimization. Only the start values have to be sorted

meeteval/wer/matching/cy_levenshtein.pyx
thequilo: No, it just gives you the line number. I added information to the error message wether reference or hypothesis is wrong

meeteval/wer/matching/cy_levenshtein.pyx
thequilo: I checked, this corresponds to the C++ `float` type. I changed it to `double`

meeteval/wer/matching/cy_levenshtein.pyx
```diff
@@ -0,0 +1,279 @@
...
+        hypothesis,
+):
+    reference, hypothesis = obj2vec(reference, hypothesis)
+
+    return levenshtein_distance_(reference, hypothesis)
+
+def levenshtein_distance_custom_cost(
+        reference,
+        hypothesis,
+        cost_del,
```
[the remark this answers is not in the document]
thequilo: You are right, the code converts the values to int, while I expected cython to not silently convert from float to int. I added a check

doc/tcpwer.md
thequilo: This is now mentioned in the section about pseudo-word-level annotation

boeddeker (approved):

PR #12, opened 2023-04-03
Add a faster C++ Implementation of MIMO WER

thequilo:
The C++ implementation in this PR is easier to read and faster than the Cython implementation. This is due to a few small optimizations getting rid of indexing operations and multiplications. The overall speedup is about 3-6 on the benchmark from the publication. This PR does not change memory usage.


thequilo:
A small correction for the runtime: The numbers from the previous message are true for my test data with 4 channels, but the speedup varies a lot with different examples and numbers of channels and can be smaller than 3.

PR #24, opened 2023-07-18
Fixes and improvements for io, cli and cpwer

boeddeker:


boeddeker:
Motivation for the name of the `parse_float` argment: `json.loads('1.1', parse_float=decimal.Decimal)`

meeteval/wer/__main__.py
```diff
@@ -444,6 +491,11 @@ def add_command(fn):
                     '--collar', type=positive_number,
                     help='Collar applied to the hypothesis timings'
                 )
+            elif name == 'regex':
```
thequilo, line 494: Does it make sense to add the option to all WERs and not just cpWER?

meeteval/wer/wer/cp.py
```diff
@@ -123,15 +124,49 @@ def cp_word_error_rate(
...
+    >>> cp_word_error_rate(r, h)
+    CPErrorRate(errors=0, length=2, insertions=0, deletions=0, substitutions=0, error_rate=0.0, missed_speaker=0, falarm_speaker=0, scored_speaker=1, assignment=(('r0', 'h0'),))
+
+    >>> cp_word_error_rate(['a b c'.split(), 'd e f'.split()], ['a b c'.split(), 'd e f'.split()])
+    CPErrorRate(errors=0, length=6, insertions=0, deletions=0, substitutions=0, error_rate=0.0, missed_speaker=0, falarm_speaker=0, scored_speaker=2, assignment=((0, 0), (1, 1)))
     """
     from meeteval.wer.matching.cy_levenshtein import levenshtein_distance
+    import meeteval.io
+
     def transcription_to_words(x):
```
thequilo, line 143: We could define a type for this, similar to what was done in the time-constrained WER

boeddeker: Maybe later, when we need that somewhere else.

meeteval/wer/__main__.py
boeddeker: Done

thequilo (approved):

PR #33, opened 2023-08-11
Allow STM input for Python interface

thequilo:


meeteval/io/ctm.py
```diff
@@ -131,6 +131,23 @@ def grouped_by_filename(self) -> Dict[str, 'CTMGroup']:
...
 
+    def to_stm(self):
+        from meeteval.io import STM, STMLine
+        stm = []
+        for filename, v1 in self.grouped_by_filename().items():
+            for speaker_id, v2 in v1.grouped_by_speaker_id().items():
+                for line in v2.lines:
+                    stm.append(STMLine(
+                        filename=filename,
+                        channel=0,
```
boeddeker, line 142: We should check, what the default for channel usually is.
Often I did it wrong. I think I also used 0, while it should be 1.

meeteval/io/keyed_text.py
```diff
@@ -50,3 +50,6 @@ def merged_transcripts(self) -> str:
     def utterance_transcripts(self) -> List[str]:
         """There is no notion of an "utterance" in CTM files."""
         raise NotImplementedError()
+
+    def filenames(self):
```
boeddeker, line 54: Add this function to the base class?

meeteval/io/stm.py
```diff
@@ -143,3 +144,63 @@ def merged_transcripts(self) -> str:
 
     def segments(self):
         return [l.segment_dict() for l in self]
+
+    def filenames(self):
+        return {l.filename for l in self}
+
+
+def iter_examples(reference: 'STM', hypothesis: 'STM', *, allowed_empty_examples_ratio=0.1):
```
boeddeker, line 152: Add a comment, that this is for internal usage?
I think we have too many functions, where the intention is internal usage, but it is not obvious.

meeteval/io/keyed_text.py
thequilo: Can we be sure that all `BaseLine` classes have a filename?

meeteval/io/stm.py
thequilo: Yeah, we also have some functions with underscores and some without. This is currently really inconsistent

meeteval/wer/__main__.py
```diff
@@ -99,7 +90,7 @@ def _load_hypothesis(hypothesis: List[Path]):
         raise RuntimeError(hypothesis, filename)
 
 
-def _load_texts(reference_paths: List[str], hypothesis_paths: List[str], regex):
+def _load_texts(reference_paths: List[str], hypothesis_paths: List[str], regex) -> (STM, List[Path], STM, List[Path]):
```
boeddeker, line 93: `(STM, List[Path], STM, List[Path])` -> `'STM | list[Path]'` (Duplicates?)

For type annotations: I prefer the newest syntax as str:
 - Modern look
 - Works in old python

meeteval/wer/__main__.py
```diff
@@ -121,56 +112,18 @@ def _glob(pathname):
...
-            keys = grouped.keys()
-            grouped = {k: v for k, v in grouped.items() if r.fullmatch(k)}
-            assert grouped, (regex, keys, 'Found nothing')
-            return grouped
+
+        def filter(stm):
+            filenames = stm.filenames()
+            filtered_filenames = [f for f in filenames if r.fullmatch(f)]
+            assert filtered_filenames, (regex, filenames, 'Found nothing')
+            return stm.filter(lambda l: l.filename in filtered_filenames)
```
boeddeker, line 123: Does the following work?
```python
        def filter(stm):
            new = stm.filter(lambda line: r.fullmatch(line.filename))
            assert len(new), (regex, stm.filenames(), 'Found nothing')
```
It's cleaner, but maybe a bit more expensive.

meeteval/io/ctm.py
thequilo: The "official" documentation (https://github.com/usnistgov/SCTK/blob/master/doc/infmts.htm) sais:

> The waveform channel.   The text of the waveform channel is not restricted by sclite.  The text can be any text string without witespace so long as the matching string is found in both the reference and hypothesis input files.

This is the only documentation I know about the STM format, so there is no official default defined. Where did you take the 1 from?

meeteval/wer/__main__.py
thequilo: This should mean `Tuple[STM, List[Path], STM, List[Path]]`. This was autofilled by PyCharm (or Copilot?), so I thought it was correct

meeteval/wer/__main__.py
thequilo: I intentionally didn't run the match on all lines since there can be many lines with the same filename. But I don't know how much of a difference this makes

meeteval/wer/wer/cp.py
```diff
@@ -157,27 +171,27 @@ def split(words):
...
         transcription_to_words(hypothesis),
-        distance_fn=levenshtein_distance,
-        siso_error_rate=_siso_error_rate,
     )
 
 
+def cp_word_error_rate_stm(reference_stm: 'STM', hypothesis_stm: 'STM') -> 'Dict[str, CPErrorRate]':
+    """
+    TODO: doc
+    TODO: return some kind of summary type?
```
boeddeker, line 189: I think a dict should be fine.
To get the summary, we may want to support `sum(cp_word_error_rate_stm().values())` (I am not sure, if I pushed that code already, probably not).

thequilo: I think `sum` already works on `ErrorRate`

meeteval/wer/wer/mimo.py
```diff
@@ -90,11 +50,11 @@ def get_keys(obj):
     )
 
     er = sum([
-        siso_word_error_rate(
-            ' '.join(r),  # Concatenate all utterances from one speaker
+        _siso_error_rate(
+            [t for r_ in r for t in r_],  # Concatenate all utterances from one speaker
```
boeddeker, line 54: `# Concatenate all utterances from one speaker` -> `# Create list with all words from one speaker`?

meeteval/wer/wer/mimo.py
```diff
@@ -107,10 +67,60 @@ def get_keys(obj):
...
+    >>> mimo_word_error_rate([['a b', 'c d'], ['e f']], ['c d', 'a b e f'])
+    MimoErrorRate(errors=0, length=6, insertions=0, deletions=0, substitutions=0, error_rate=0.0, assignment=[(0, 1), (1, 1), (0, 0)])
+
+    >>> mimo_word_error_rate({'A': ['a b', 'c d'], 'B': ['e f']},
+    ...                      {'O1': 'c d', 'O2': 'a b e f'})
+    MimoErrorRate(errors=0, length=6, insertions=0, deletions=0, substitutions=0, error_rate=0.0, assignment=[('A', 'O2'), ('B', 'O2'), ('A', 'O1')])
+
+    """
+    if isinstance(reference, STM) or isinstance(hypothesis, STM):
+        assert isinstance(hypothesis, STM) and isinstance(reference, STM)
```
boeddeker, line 94: `assert isinstance(hypothesis, STM) and isinstance(reference, STM), (type(hypothesis), type(reference))`

meeteval/wer/wer/orc.py
```diff
@@ -115,8 +74,8 @@ def orc_word_error_rate(
     )
 
     er = sum([
-        siso_word_error_rate(
-            ' '.join(r),  # Concatenate all utterances from one speaker
+        _siso_error_rate(
+            [t for r_ in r for t in r_],  # Concatenate all utterances from one speaker
```
boeddeker, line 78: `# Concatenate all utterances from one speaker` -> `# Create list with all words from one speaker`?

meeteval/wer/wer/siso.py
```diff
@@ -42,12 +45,31 @@ def siso_word_error_rate(
     >>> siso_word_error_rate(reference='This is wikipedia', hypothesis='This wikipedia')  # Deletion example from https://en.wikipedia.org/wiki/Word_error_rate
     ErrorRate(errors=1, length=3, insertions=0, deletions=1, substitutions=0, error_rate=0.3333333333333333)
     """
+    if isinstance(reference, (KeyedText, STM)) or isinstance(hypothesis, (KeyedText, STM)):
```
boeddeker, line 48: It is strange that this function accepts STM and KeyedText.
Could you give a hint, where this is helpful?

boeddeker:
PR motivation:

CLI supports a common input format for each WER type.
The python code uses the WER specific input type (i.e. the natural input format for each WER type, e.g. for cpWER it is one transcript for each speaker without timing information, while tcpWER expect words or short segments with timing information).
It is not that easy to get the input format with the first try correct (Especially ORC, MIMI and tcp).

Now: With the STM interface, each WER has a high level interface and STM is well defined.

meeteval/wer/wer/mimo.py
thequilo: It's not necessarily words at this point. The function accepts anything, so it could be characters here

meeteval/wer/wer/siso.py
thequilo: I'm not sure if this function should accept STM. The command line interface doesn't accept STM. So, I might remove STM

boeddeker:
Took some time to find it.
 - On https://github.com/nryant/dscore I found `channel (1-indexed) that turn is on; should always be 1`. (Currently, the best argument to select a default, because it is the only argument)
 - From the RWTH I saw an STM file and there was the default 1.
 - In ESPnet I saw some rttm files and they have as default 1.

While that is not a strong argument, it is at least an argument to select one over another.

meeteval/io/ctm.py
boeddeker: Took some time to find it.
 - On https://github.com/nryant/dscore I found `channel (1-indexed) that turn is on; should always be 1`. (Currently, the best argument to select a default, because it is the only argument)
 - From the RWTH I saw an STM file and there was the default 1.
 - In ESPnet I saw some rttm files and they have as default 1.

While that is not a strong argument, it is at least an argument to select one over another.

meeteval/io/keyed_text.py
boeddeker: I think all classes have the concept of filename, otherwise you cannot find the mapping between ref and hyp

meeteval/wer/__main__.py
boeddeker: Ah, ok. I thought the code only translated from dict to stm.

meeteval/wer/wer/cp.py
boeddeker: Ok, then I would return a dict. When the user wants to have the summary, it is easy to use sum.
We could give a hint for this in the doctext.

meeteval/wer/wer/mimo.py
boeddeker: Yes, I agree, that it could be anything that is hashable. But I didn't know, how to phrase it, so I used the most common case and that is: word.

meeteval/wer/wer/siso.py
boeddeker: Yes, from the command line, it is clear, what to do with an STM:
 - SISO-WER on segments, like it is done in CHiME-5
 - SISO-WER on concatenated transcripts (cpWER without perm solver).
 - Assert that the above two yield the same WER, i.e. for each file only one transcript.

meeteval/io/keyed_text.py
thequilo: Done

meeteval/wer/__main__.py
thequilo: Changed it to `Tuple[STM, List[Path], STM, List[Path]]`.

meeteval/wer/wer/error_rate.py
```diff
@@ -58,7 +58,7 @@ def __radd__(self, other: 'int') -> 'ErrorRate':
     def __add__(self, other: 'ErrorRate') -> 'ErrorRate':
         """Combines two error rates"""
         if not isinstance(other, ErrorRate):
-            raise NotImplemented
+            raise NotImplementedError()
```
boeddeker, line 61: Can you change this to `return NotImplemented`?

thequilo: Done

boeddeker (approved):
Are the functions for the end users accessible in the `meeteval.wer` namespace?

thequilo:
They are accessible via, e.g., `meeteval.wer.wer.cp.cp_word_error_rate_stm`. But not directly under `meeteval.wer`. I'll add the import there

PR #39, opened 2023-10-05
Add options for sorting and track self-overlap

thequilo:


.github/workflows/pytest.yml
```diff
@@ -44,4 +45,4 @@ jobs:
         python -m flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
     - name: Test with pytest
       run: |
-        python -m pytest
+        python -m pytest --doctest-modules
```
boeddeker, line 48: Could you use a pytest.ini as in paderbox=

meeteval/wer/matching/cy_levenshtein.pyx
```diff
@@ -113,59 +111,9 @@ def levenshtein_distance(
...
-            raise ValueError(
-                f'The start times of the annotations must be increasing, which they are not for the hypothesis. '
-                f'(found at least one interval where {t.first} < {last_start}) '
-                f'This might be caused by overlapping segments, see the (potential) previous warning.'
-            )
-        last_start = t.first
 
-    return reference, hypothesis, reference_timing, hypothesis_timing
+cdef _timing_is_integer(timing):
+    return all(isinstance(t[0], int) and isinstance(t[1], int) for t in timing)
```
boeddeker, line 116: What is the usual result? If it is True, a list comprehension is faster, otherwise the generator comprehension is faster. Or is it different in Cython?

meeteval/wer/wer/error_rate.py
```diff
@@ -2,6 +2,52 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class SelfOverlap:
+    """
+    This class represents the self-overlap of a reference or a hypothesis
+    """
+    overlap_time: float
+    total_time: float
+
+    overlap_rate: float = dataclasses.field(init=False)
```
boeddeker, line 17: Put this also at the first position?

meeteval/wer/wer/time_constrained.py
```diff
@@ -153,20 +159,26 @@ class TimeMarkedTranscript:
...
         return False
 
+    def get_self_overlap(self):
+        last_end = 0
+        self_overlap = 0
+        total = 0
+        for t in sorted(self.timings):
+            if last_end > t[0]:
+                # TODO: Is this the best way to measure self-overlap?
+                self_overlap += 2 * (last_end - t[0])
```
boeddeker, line 177: Could you add a small test for this function?
I would count the self-overlap only once.

There may be two bugs, could you check:
 -  What happens, when you have two back channels? Here, an example activity pattern:
```
#########################
   ######       
                               ######
 -----------------------> time axis
```
 - What happens for triple overlap?
```
###############
##########
      ###########
```

meeteval/wer/wer/time_constrained.py
```diff
@@ -239,10 +248,13 @@ def _repr_pretty_(self, p, cycle):
 TimeMarkedTranscriptLike = 'TimeMarkedTranscript | STM | List[Segment]'
 
 
+def apply_collar(s: TimeMarkedTranscript, collar: float):
+    return replace(s, timings=[(max(t[0] - collar, 0), t[1] + collar) for t in s.timings])
```
boeddeker, line 252: Is negative time not allowed?

meeteval/wer/wer/time_constrained.py
```diff
@@ -370,12 +381,53 @@ def sort_segments(s: TimeMarkedTranscript):
...
+            f'{len(segments.transcript)} != {len(segments.timings)}'
+        )
+
+    for t in segments.timings:
+        if t[1] < t[0]:
+            raise ValueError(f'The end time of an interval must be larger than the start time. Found {t} in {name}')
+
+    prune = True
+
+    assert sort in (True, False, 'segment', 'word')
```
boeddeker, line 398: `assert sort in (True, False, 'segment', 'word'), sort`

meeteval/wer/wer/time_constrained.py
```diff
@@ -370,12 +381,53 @@ def sort_segments(s: TimeMarkedTranscript):
...
+    if sort in (True, 'segment', 'word'):
+        segments = sort_segments(segments)
+
+    words = get_pseudo_word_level_timings(segments, pseudo_word_level_timing)
+
+    # Check whether words are sorted by start time
+    words_sorted = sort_segments(words)
+    if words_sorted != words:
```
boeddeker, line 406: Should this be in a `elif` after the `if sort == 'word':`?

meeteval/wer/wer/time_constrained.py
```diff
@@ -386,27 +438,46 @@ def time_constrained_siso_word_error_rate(
...
+    hypothesis = apply_collar(hypothesis, collar)
 
-    return _time_constrained_siso_error_rate(
-        reference.transcript, hypothesis.transcript,
-        reference.timings, hypothesis.timings,
+    er = _time_constrained_siso_error_rate(
+        reference_.transcript, hypothesis_.transcript,
+        reference_.timings, hypothesis_.timings,
     )
+    er = replace(
```
boeddeker, line 465: Add a comment like: `pseudo_word_level_timing and collar change the time stamps, hence calculate the overlap with the original time stamps`

meeteval/wer/matching/cy_levenshtein.pyx
```diff
@@ -176,18 +124,17 @@ def time_constrained_levenshtein_distance(
         cost_ins: uint = 1,
         cost_sub: uint = 1,
         cost_cor: uint = 0,
+        prune: bool = True
```
boeddeker, line 127: What does `time_constrained_levenshtein_distance_v2_`, when prune is True, while it should be False?
If the result would be wrong, I suggest two changes:
 - Change the default to False or None (None would mean calculate, whether prune can be used, not sure about the overhead)
 - Add a comment, that no user should set prune to True, it will be computed.

tests/test_time_constrained.py
```diff
@@ -180,3 +182,59 @@ def test_tcpwer_input_formats():
...
+    r2 = TimeMarkedTranscript(['a b c d e f g h'], [(0, 6)])
+    er = time_constrained_minimum_permutation_word_error_rate(
+        [r1], [r2], reference_sort='word',
+    )
+    assert er.error_rate == 0.75
+
+    er = time_constrained_minimum_permutation_word_error_rate(
+        [r1], [r2], reference_sort='segment',
+    )
+    assert er.error_rate == 0.75
```
boeddeker, line 215: Add a hint, why `reference_sort='segment'` doesn't change the WER here?

tests/test_time_constrained.py
```diff
@@ -180,3 +182,59 @@ def test_tcpwer_input_formats():
...
+    )
+    assert er.error_rate == 0.25
+
+    # False means the user provides the sorting, so we can pass anything
+    r1 = TimeMarkedTranscript(['e f g h', 'a b c d'], [(4, 7), (0, 3)])
+    r2 = TimeMarkedTranscript(['a b c d e f g h'], [(0, 7)])
+    er = time_constrained_minimum_permutation_word_error_rate(
+        [r1], [r2], reference_sort='segment',
+    )
+    assert er.error_rate == 0.5
```
boeddeker, line 235: I don't get, why the error rate is 0.5.

meeteval/wer/matching/cy_levenshtein.pyx
thequilo: It's usually `False` when processing STMs since they usually use (floating-point) seconds as a unit. I don't know how this is translated to C++ by Cython, but I believe this still uses the Python functions.

meeteval/wer/wer/error_rate.py
thequilo: Done

.github/workflows/pytest.yml
thequilo: Done

meeteval/wer/wer/time_constrained.py
thequilo: You are right, negative time is currently allowed. I copied this to keep the old behavior. An old implementation used unsigned types. I'll remove the max operation

thequilo: Done

meeteval/wer/wer/time_constrained.py
thequilo: I actually wanted to print this warning in all cases. Even if `sort == 'word'` the warning gives you a hint whether the sorting changes the order or not

meeteval/wer/wer/time_constrained.py
thequilo: Done

meeteval/wer/matching/cy_levenshtein.pyx
thequilo: Added a function that computes whether pruning is applicable and set the default to `'auto'`

meeteval/wer/matching/cy_levenshtein.pyx
boeddeker: Ok, then the generator comprehension is fine.

meeteval/wer/wer/time_constrained.py
boeddeker: Yes, you are right.

tests/test_time_constrained.py
thequilo: Agreed, it's hard to see because the timestamps result in numbers with many decimal points, and no collar is set. I changed the time stamps so that the error rate becomes 0

meeteval/wer/wer/time_constrained.py
thequilo: Done

meeteval/wer/wer/time_constrained.py
thequilo: I opted to count the overlap twice because it seems wrong to have timestamps `(0, 2)` and `(0, 2)` but 50% overlap was reported. The intuition could be: Count for how many samples at least one other sample is active at the same time. But as the comment says, I'm not sure whether this is the best way to do it.

Triple overlap is currently not handled correctly, I'll fix that

meeteval/wer/wer/time_constrained.py
```diff
@@ -370,12 +381,53 @@ def sort_segments(s: TimeMarkedTranscript):
...
+    if sort in (True, 'segment', 'word'):
+        segments = sort_segments(segments)
+
+    words = get_pseudo_word_level_timings(segments, pseudo_word_level_timing)
+
+    # Check whether words are sorted by start time
+    words_sorted = sort_segments(words)
+    if words_sorted != words:
+        if sort is not True:
+            logger.warning(f'The order of word-level timings contradicts the segment-level order in {name}')
```
boeddeker, line 408: How about changing the warning to
```
contradictions = [a!=b for a, b in zip(words_sorted.transcript, words.transcript)]
f'The order of word-level timings contradicts the segment-level order in {name}: {sum(contradictions)} of {len(contradictions)} times.'
```
Assuming the contradictions are local, this will give a hint, how often that happens.

meeteval/wer/wer/time_constrained.py
thequilo: Analternative: Count overlap only once in the total time (which is then smaller than the sum of the segment times) and count overlap once

boeddeker: I find it more natural, when self_overlap would be 2 seconds and the total time would be 2 seconds for your example  `(0, 2) and (0, 2)`. In that way, the overlap cannot exceed the duration of the recording with only double overlap.
The implementation could be: Count the overlap without the factor 2 and substract the overlap at the end from the total time. This would automatically handle triple overlap.

tests/test_time_constrained.py
boeddeker: Yes, it would be good to have extreme cases, such that it is obvious, when two words don't overlap.

meeteval/wer/wer/time_constrained.py
thequilo: Done

meeteval/wer/matching/cy_levenshtein.pyx
```diff
@@ -113,132 +121,114 @@ def levenshtein_distance(
...
+
+    # Convert to vectors of integers for the C++ implementation
     reference, hypothesis = obj2vec(reference, hypothesis)
 
-    args = (reference, hypothesis,
+    # Dispatch to the correct implementation
+    if prune == 'auto':
+        prune = _timings_sorted(reference_timing) and _timings_sorted(hypothesis_timing)
+        logging.debug(
+            f'prune is set to "auto". '
```
boeddeker, line 168: `f'Changed prune from "auto" to {prune!r}. '`

meeteval/wer/wer/time_constrained.py
thequilo: They are not guaranteed to be local. In case of a back-channel scenario this number could become large although only a single word is in the wrong position

meeteval/wer/matching/cy_levenshtein.pyx
thequilo: I feel that your message could be understood wrongly. I shortened the message

meeteval/wer/wer/time_constrained.py
boeddeker: With local I mean, that even in those cases the number should be small. e.g. LibriCSS eval has 11159 Words, and since it should rarely happen, that something went wrong, a report of `100 of 11159` vs `5000 of 11159` would help the user, that in the latter case something went wrong.

thequilo: Okay, I tested your suggestion on the Chime7 baseline data and the numbers seem reasonable.

boeddeker (approved):

PR #43, opened 2023-12-05
add der md_eval_22

boeddeker:


meeteval/der/md_eval.py
```diff
@@ -0,0 +1,175 @@
...
+import dataclasses
+from pathlib import Path
+
+import meeteval.io
+from meeteval.wer.wer.error_rate import ErrorRate
+
+
+def _fix_channel(r):
+    return meeteval.io.rttm.RTTM([
+        line.replace(channel='1')  # Thilo puts there usually some random value, e.g. <NA> for hyp and 0 for ref, while common default is 1
```
thequilo, line 16: Can you change the default in the `RTTM` class? 

I think the comment should say that the original (?) RTTM definition (Appendix A in "The 2009 (RT-09) Rich Transcription Meeting Recognition Evaluation Plan") doesn't say anything about the channel format or defaults, but dscore enforces a "1" (https://github.com/nryant/dscore#rttm)

meeteval/der/md_eval.py
```diff
@@ -0,0 +1,175 @@
...
+        for line in r
+    ])
+
+
+@dataclasses.dataclass(frozen=True)
+class DiaErrorRate:
+    """
+
+    """
+    error_rate: 'float | decimal.Decimal'
```
thequilo, line 26: Why is this not computed from the times? The WER `ErrorRate` has `dataclasses.field(init=False)` and always computes it in the `__post_init__`

meeteval/der/md_eval.py
```diff
@@ -0,0 +1,175 @@
...
+        return DiaErrorRate(
+            error_rate=None,
+            scored_speaker_time=self.scored_speaker_time + other.scored_speaker_time,
+            missed_speaker_time=self.missed_speaker_time + other.missed_speaker_time,
+            falarm_speaker_time=self.falarm_speaker_time + other.falarm_speaker_time,
+            speaker_error_time=self.speaker_error_time + other.speaker_error_time,
+        )
+
+
+def md_eval_22(
```
thequilo, line 70: Can you split this into a Python function that takes `RTTM` objects and the cli function that takes file paths?

For WERs, the cli functions are defined in `__main__.py`.

meeteval/der/md_eval.py
```diff
@@ -0,0 +1,175 @@
...
+        deletions, = re.findall(r'MISSED SPEAKER TIME =([\d.]+) secs', cp.stdout)
+        insertions, = re.findall(r'FALARM SPEAKER TIME =([\d.]+) secs', cp.stdout)
+        substitutions, = re.findall(r'SPEAKER ERROR TIME =([\d.]+) secs', cp.stdout)
+
+        def convert(string):
+            return decimal.Decimal(string)
+
+        details = dict(
+            scored_speaker_time=convert(length),
+            # errors=float(errors),
```
thequilo, line 139: Remove this line?

meeteval/der/md_eval.py
```diff
@@ -0,0 +1,175 @@
...
+            missed_speaker_time=convert(deletions),
+            falarm_speaker_time=convert(insertions),
+            speaker_error_time=convert(substitutions),
+            error_rate=convert(error_rate) / 100,
+        )
+        return DiaErrorRate(**details)
```
thequilo, line 145: Why don't you construct the `DiaErrorRate` directly? The `details` dict seems redundant

meeteval/der/md_eval.py
```diff
@@ -0,0 +1,175 @@
...
+                f'({summary.error_rate})\n'
+                f'does not match the average error rate of md-eval-22.pl '
+                f'applied to each recording ({md_eval.error_rate}).'
+            )
+
+    from meeteval.wer.__main__ import _save_results
+    _save_results(per_reco, hypothesis_paths, per_reco_out, average_out)
+
+
+if __name__ == '__main__':
```
thequilo, line 173: Why is this necessary? Isn't `meeteval.der.__main__` enough for CLI?

meeteval/wer/__main__.py
```diff
@@ -337,8 +375,111 @@ def positive_number(x: str):
 
         return x
 
-    def add_command(fn):
-        command_parser = commands.add_parser(
+    def add_argument(self, command_parser, name, p):
```
thequilo, line 378: Rename `add_argument` -> `add_arguments`?

meeteval/wer/__main__.py
```diff
@@ -312,20 +345,25 @@ def average(files, out):
...
-    parser = argparse.ArgumentParser()
-    parser.add_argument('--version', action='store_true', help='Show version')
-
-    # Logging and verbosity
-    logging.addLevelName(100, 'SILENT')     # Add a level that creates no output
-    parser.add_argument(
-        '--log-level', help='Log level', default='WARNING',
-        choices=['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL', 'SILENT']
-    )
+class CLI:
```
thequilo, line 348: Does it make sense to move this class somewhere else and import it in all `__main__.py` files? It seems weird that the DER CLI script imports the WER CLI script. `meeteval.cli`?

Same applies to the other functions that are imported in the DER CLI.

meeteval/der/md_eval.py
boeddeker: While the WERs are calculated from our code, this is the values that is computed by md_eval_22.pl.
I wanted to be as close as possible to the original. (except, that it is here just the ratio and not the ratio in percent)

meeteval/wer/__main__.py
boeddeker: While the CLI class can be used for WER and DER, it is mainly designed for WER.
Since we don't know, how future interfaces look like, I thought, we use here the concept that is used in kaldi:
Define the function where it is used and import it, where it also works.

But I could also move the code to cli.py

meeteval/der/md_eval.py
boeddeker: That was my old code. removed

meeteval/der/md_eval.py
boeddeker: Done and added your comment.

meeteval/der/md_eval.py
boeddeker: done

meeteval/der/md_eval.py
boeddeker: done

meeteval/wer/__main__.py
boeddeker: That function processes only one argument. That makes it easier in the future to add new arguments or disable arguments in the future in subclasses.

PR #44, opened 2023-12-07
Add SegLST format

thequilo:
Adds the `SegLST` (Segment-wise Long-form Speech Transcription Annotation) format. This format has been used in past CHiME challenges. It becomes the default for all WER functions as it is the most flexible format.


meeteval/io/seglst.py
```diff
@@ -0,0 +1,388 @@
...
+    speaker: str
+    segment_index: int
+
+    # Unused but by MeetEval but present in some file formats. They are defined here for compatibility and
+    # conversion in both directions
+    channel: int
+    confidence: float
+
+
+class SegLSTMixin:
```
boeddeker, line 30: This should be defined in the baseclass.

meeteval/io/seglst.py
```diff
@@ -0,0 +1,388 @@
...
+
+    >>> @seglst_map(required_keys=('speaker',))
+    ... def fn(seglst, *, speaker='X'):
+    ...     return seglst.map(lambda x: {**x, 'speaker': speaker})
+    >>> from meeteval.io.stm import STM
+    >>> fn(STM.parse('X 1 A 0 1 a b c'))
+    STM(lines=[STMLine(filename='X', channel=1, speaker_id='X', begin_time=0, end_time=1, transcript='a b c')])
+    >>> from meeteval.io.rttm import RTTM
+    >>> fn(RTTM.parse('SPEAKER CMU_20020319-1400_d01_NONE 1 130.430000 2.350 <NA> <NA> juliet <NA> <NA>'))
+    RTTM(lines=[RTTMLine(type='SPEAKER', filename='CMU_20020319-1400_d01_NONE', channel='1', begin_time=130.43, duration=2.3499999999999943, othography='<NA>', speaker_type='<NA>', speaker_id='X', confidence='<NA>', signal_look_ahead_time='<NA>')])
```
boeddeker, line 296: We should switch to decimal as default representation for numbers to avoid:
 - Loss of visual precision: `130.430000` vs `130.43`
 - Issue that float has base 2: `2.350` vs `2.3499999999999943`

thequilo: Done

meeteval/io/stm.py
```diff
@@ -1,14 +1,9 @@
-import sys
-import typing
 from dataclasses import dataclass
-from typing import List, NamedTuple
+from typing import List, Self
```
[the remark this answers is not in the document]
boeddeker: `List` can be replaced by `list`.

meeteval/io/base.py
```diff
@@ -20,12 +23,33 @@
...
 
 
+
+class BaseABC:
+    @classmethod
+    def new(cls, d, **defaults):
+        # Example code:
+        # from meeteval.io.seglst import asseglst
+        # seglst = asseglst(d).map(lambda s: {**s, **defaults})
+        # ... (conert seglst to cls)
```
boeddeker, line 33: *convert

meeteval/io/base.py
```diff
@@ -20,12 +23,33 @@
...
+
+
 @dataclass(frozen=True)
 class BaseLine:
     @classmethod
     def parse(cls, line: str) -> 'Self':
         raise NotImplementedError(cls)
 
+    @classmethod
+    def from_seglst(cls, segment: 'SegLstSegment') -> 'Self':
```
boeddeker, line 47: Change the name to `from_dict`? The input is an entry from a seglst, which is a dict with special keys.

thequilo: Done

meeteval/io/base.py
```diff
@@ -20,12 +23,33 @@
...
 class BaseLine:
     @classmethod
     def parse(cls, line: str) -> 'Self':
         raise NotImplementedError(cls)
 
+    @classmethod
+    def from_dict(cls, segment: 'SegLstSegment') -> 'Self':
+        raise NotImplementedError(cls)
+
+    def to_seglst(self) -> 'SegLstSegment':
```
boeddeker, line 50: Same issue as `from_seglst`, but I am nor sure, whether `to_dict` would be a good name.
It is too close to `dataclasses.asdict`.

meeteval/io/py.py
```diff
@@ -0,0 +1,254 @@
...
+
+def _invert_python_structure(t: 'SegLST', types, keys):
+    if len(types) != len(keys):
+        if len(types) < len(keys):
+            # _convert_python_structure adds a dummy key if the final key is not set. We have to remove it here
+            # again.
+            for k in keys[len(types) - 1:-1]:
+                if len(t.unique(k)) != 1:
+                    raise ValueError(
+                        f'Cannot convert SegLST to Tidy with t.keys={t.keys!r}, types={types!r} and keys={keys!r}. '
```
boeddeker, line 90: Old name `Tidy`

meeteval/io/py.py
```diff
@@ -0,0 +1,254 @@
...
+        'a b c'
+        >>> convert_cycle(['a b c', 'd e f'], keys=('speaker',))
+        ['a b c', 'd e f']
+        >>> convert_cycle({'A': 'a b c', 'B': 'd e f'}, keys=('speaker',))
+        {'A': 'a b c', 'B': 'd e f'}
+        >>> s = NestedStructure({'B': 'd e f', 'A': 'a b c'}, level_keys=('speaker',))
+        >>> s2 = NestedStructure(['a b c', 'd e f'], level_keys=('speaker',))
+        >>> s.new(s2.to_seglst()).structure
+        {0: 'a b c', 1: 'd e f'}
+        >>> s2.new(s.to_seglst()).structure   # Keys are sorted when converting to list (TODO: is this expected behavior?)
```
boeddeker, line 168: Since python 3.7 the order of dicts is part of the language (implemented since cpython 3.6). We shouldn't touch the order.
I see only reasons, there the sort is annoying, but no reason, where it improves the code.

meeteval/io/seglst.py
```diff
@@ -0,0 +1,388 @@
...
+    A collection of segments in SegLST format. This the input type to most
+    functions in MeetEval that process transcript segments.
+    """
+    segments: 'List[SegLstSegment]'
+
+    # Caches
+    _unique = None
+
+    @cached_property
+    def keys(self):
```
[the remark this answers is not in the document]
boeddeker: How about something like the following?
```python
@property
class T:
    def __init__(self, outer):
        self._outer = outer
    def keys(self):
        return set.intersection(*[set(s.keys()) for s in self._outer.segments])
    def __getitem__(self, key):
        return [s[key] for s in self._outer.segments]
```
Do you have a suggestion for an alternative name? (I took `T` from numpy, where it means transpose)

```python
t = asseglst(['a b c', 'd e f', 'g h i'])
print(t.T.keys())  # 'words', 'segment_index', 'speaker'
print(t.T['words'])  # ['a b c', 'd e f', 'g h i']
```

meeteval/wer/wer/time_constrained.py
```diff
@@ -378,87 +416,153 @@ def remove_overlaps(
...
-        segments = sort_segments(segments)
+        segments = segments.sorted('start_time')
 
     words = get_pseudo_word_level_timings(segments, pseudo_word_level_timing)
 
     # Check whether words are sorted by start time
-    words_sorted = sort_segments(words)
-    prune = True
+    words_sorted = words.sorted('start_time')
+    # TODO: only check relevant keys? That would speed things up if the user provides a lot of custom keys
```
boeddeker, line 486: That is not necessary:
```
class Foo:
    def __init__(self, obj):
        self.obj = obj
    def __eq__(self, other):
        print('__eq__')
        return self.obj == other.obj

a = [Foo('a'), Foo('b')]
b = [Foo('a'), Foo('b')]

print(a == a)  # True
print(a == b)  # __eq__ __eq__ True
```
The identity of the objects should be preserved by sort and hence the comparison is cheap, if they are the same.

You may want to add a comment and change the list comprehension:
```
contradictions = [not(a is b or a == b) for a, b in zip(words_sorted, words)]
```

meeteval/io/py.py
thequilo: The reason is that if you change the keys, you want the order to change when you convert back to list. Example:

```Python
d = ['abc', 'def']
c = asseglistconvertible(d)
s = c.to_seglst()
s[0]['speaker'] = 1
s[1]['speaker'] = 0
c.new(s)
# Should this be
['abc', 'def']
# or 
['def', 'abc']
# ?
```

meeteval/io/py.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Empty nested structures are allowed but not represented in the `SegLST` format (TODO)
+        >>> NestedStructure([['ab'], []], level_keys=('speaker', 'segment_index')).to_seglst()
+        SegLST(segments=[{'words': 'ab', 'segment_index': 0, 'speaker': 0}])
+
+        The last key is filled with a dummy key
+        >>> NestedStructure(['a b c', 'd e f'], level_keys=('speaker', 'channel')).to_seglst()
+        SegLST(segments=[{'words': 'a b c', 'channel': 0, 'speaker': 0}, {'words': 'd e f', 'channel': 0, 'speaker': 1}])
+
+        Nested structures beyond the specified groups are by default not allowed. With `ensure_word_is_string=False`,
```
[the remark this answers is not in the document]
thequilo: An example would be a generalized `apply_assignment` where you can input, for example `[SomeObject(), SomeObject()]`. But currently it's unused and should not be used directly by the user (I think) because it can lead to confusing behavior

meeteval/io/seglst.py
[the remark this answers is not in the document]
thequilo: I like `t.T['words']` a lot. This operation is frequently used in the codebase but is currently always implemented manually with a list comprehension. From pandas we could use the name `columns`, but I think I like `T` better.

meeteval/io/seglst.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def map(self, fn):
+        """
+        Applies `fn` to all segments and returns a new `SegLST` object with the results.
+        """
+        return SegLST([fn(s) for s in self.segments])
+
+    def flatmap(self, fn):
+        """
+        Applies `fn` to all segments, flattens the output and returns a new `SegLST` object with the results.
```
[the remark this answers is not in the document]
thequilo: I thought it was a well-known and commonly used operation. I'll update the docstring to be clearer about what the function is doing and give a simple example for splitting utterances into words:

```python
s = SegLST([{'words': 'a b'}, {'words': 'c d'}])
s.flatmap(lambda x: [{'words': w} for w in x['words'].split()])
SegLST(segments=[{'words': 'a'}, {'words': 'b'}, {'words': 'c'}, {'words': 'd'}])
```

meeteval/io/py.py
boeddeker: Hmm. So with and without sort, the result can be unexpected:
 - `speaker` $\in$ `{'spk10,spk200,...'}`, if we sort them, then we have to use natsort
 - `speaker` $\in$ `{0, 1, ...}` the user expects a sort
 - `speaker` $\in$ `{'0', '1', ..., '10', ...}` should this be sorted? If yes, we have to use natsort

Do we need the support for unnamed speakers? It makes everything much more complicated.

meeteval/io/py.py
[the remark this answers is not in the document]
boeddeker: So it is already used in some code?

At the moment, the input may allow too much.
The `final_types` argument may be difficult to use, because it works only on a finite set of the input structure.

meeteval/io/seglst.py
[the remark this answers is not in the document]
boeddeker: I now realized, that keys is a property and not a function. With `cached_property` and caching inside the class, the second call could get a significant speedup. How likely is it, that a user manipulates a SegLST, while it should be treated as immutable?

meeteval/io/seglst.py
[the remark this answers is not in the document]
boeddeker: It is rarely used in python. I usually see list comprehensions, where a flatmap could be used.
The code is longer, but easier to modify (i.e. not restricted to the flatmap operation).
```python
SegLST([
    entry
    for x in s
    for entry in [{'words': w} for w in x['words'].split()]
])
```

meeteval/io/stm.py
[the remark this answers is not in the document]
thequilo: Done. I also replaced `Tuple` and `Dict`

meeteval/io/py.py
thequilo: Done

meeteval/io/py.py
```diff
@@ -0,0 +1,254 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class NestedStructure(BaseABC):
+    """
+    Wraps a Python structure where the structure levels represent keys.
+
+    Example structure for cpWER:
+        ```python
+        structure = {
+            'Alice': ['segment1', 'segment2'],   # Speaker 1: Alice
```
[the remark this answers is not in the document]
thequilo: Done

meeteval/io/py.py
thequilo: We need support for converting them to `SegLST` as not to break code that already uses the Python interface. The inversion is not important.

We could forbid non-int keys for conversion to lists with a good error message

meeteval/io/py.py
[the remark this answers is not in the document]
thequilo: It is currently unused. I had a draft of a generalized apply_assignment function, but that turned out to be too complicated to push so I removed it. I can remove the `final_types` arg as long as it is not needed

meeteval/wer/wer/time_constrained.py
```diff
@@ -351,24 +389,24 @@ def get_pseudo_word_level_timings(
...
+@seglst_map()
 def remove_overlaps(
-        s: TimeMarkedTranscript,
+        t: SegLST,
         max_overlap: float = 0.4,
         warn_message: str = None,
-) -> TimeMarkedTranscript:
+) -> SegLST:
     """
     Remove overlaps between words or segments in a transcript.
```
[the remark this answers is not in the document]
thequilo: It it reasonable to sort in the function? I think that's the better solution.

meeteval/wer/wer/time_constrained.py
thequilo: Thanks, I missed that. The critical point is the  `sorted_words != words` line, but I the same thing should hold there. I'll just remove the TODO comment

meeteval/io/base.py
thequilo: That's the reason why I didn't rename it yet. With the type annotation `def to_dict(self) -> 'SegLstSegment'` could be good enough, but it sounds too generic / too close to `dataclasses.asdict`

meeteval/io/seglst.py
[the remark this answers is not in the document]
thequilo: How about this?

```python
def apply(self, fn):
    return SegLST(fn(self.segments))
```

Would result in 

```python
s.apply(lambda s: [
    entry
    for x in s
    for entry in [{'words': w} for w in x['words'].split()]
])
```

Now that I look at it, it doesn't improve over the list-comprehension variant. (Python syntax for this kind of thing sucks)

meeteval/io/py.py
boeddeker: Yes, that is a good idea.

meeteval/io/py.py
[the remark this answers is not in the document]
boeddeker: We could rename that argument to `_final_types` (i.e. make it private) and make a statement, that it is WIP for a generalized apply_assignment function.
It may turn out later, that we want to implement it differently to what `final_types` does.

In general:
When we allow less, we have later more freedom to change something to add new features.

meeteval/io/seglst.py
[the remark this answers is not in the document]
boeddeker: So the `cached_property` on this keys method may introduce bugs?

Ok, then, no caching sound better.

Maybe we should add a comment to that class, that it should be treated as immutable. I am not sure, whether we will hit a performance issue without caching. (In lazy_dataset I had to add code to cache the keys, because it was called too often. I guess here it will not be the case.)

meeteval/io/seglst.py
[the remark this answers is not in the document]
boeddeker: Yes, working with wrappers around list, it is annoying and error-prone to work with list comprehensions.

I like the `flatmap` code more. It solves this particular problem better and the name is well-defined by java.

Since I am more familiar with data loading pipelines, the description as map followed by unbatch is more familiar to me.

The `apply` code gets shorter, when you don't use an `apply` function:
```python
s.__class__([
    entry
    for x in s
    for entry in [{'words': w} for w in x['words'].split()]
])
```

meeteval/wer/wer/time_constrained.py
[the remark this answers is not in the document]
boeddeker: I think the initial `TimeMarkedTranscript` was sorted. Now I am not sure, whether the input will be sorted.

> It it reasonable to sort in the function? I think that's the better solution.

It depends. When the sort is not necessary, it may affect the runtime.
When I tried to optimize the tcpWER by mapping floats to ints, the sort was far too expensive to give a runtime improvement.

meeteval/wer/wer/time_constrained.py
boeddeker: The `words_sorted != words` will be cheap, because python first tests the identity and in most cases it will be True. So that should be fine.

A problem could be the `a != b`, when a warning is raised (error is fine, that one should never be caught).
In that case, we could hit the case that you mentioned.

You could leave there a comment. I was not sure, whether Python uses first an identity check on the elements of a list. Depending on the outer code, this could change in the future (identity fails, while equality holds).

meeteval/io/base.py
boeddeker: Any suggestion for an alternative name? Or we may use the length `to_segLstSegment`

meeteval/io/py.py
thequilo: Done

meeteval/wer/wer/time_constrained.py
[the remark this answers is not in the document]
thequilo: The sort is necessary and the function is currently unused. It because unused when we added the sort arguments to the tcpWER. I would like to keep the function because it is needed when we want to include an asclite wrapper that evaluates in the same way as the original Libri-CSS papers.

meeteval/io/py.py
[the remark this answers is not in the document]
thequilo: Okay, I made it private

meeteval/io/seglst.py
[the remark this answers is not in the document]
thequilo: Okay, I removed the caching. We can keep an eye on this and add some sort of caching when we notice performance issues

meeteval/wer/wer/time_constrained.py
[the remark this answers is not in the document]
boeddeker: Ah, ok. Then adding the sort is a good idea.

meeteval/io/seglst.py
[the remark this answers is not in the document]
thequilo: Updated the type annotations and docstring to make clearer, what `flatmap` does

meeteval/wer/wer/time_constrained.py
thequilo: Python's sort doesn't keep identity when the list is already sorted, it only keeps the identity of the elements:
```py
l = [1,2,3]
sorted(l) is l
Out[3]: False
sorted(l) == l
Out[4]: True

l = [{'k': 0}, {'k': 1}]
sorted(l, key=lambda x: x['k'])[0] is l[0]
Out[8]: True
```
But that shouldn't hurt much either. I'll add a comment

meeteval/io/base.py
thequilo: Maybe `to_seg_lst_segment`? But that is so long. I think I prefer `to_dict` here

meeteval/wer/wer/time_constrained.py
boeddeker: Yes, that was what I expected: Python shouldn't modify the entries.
So whatever we do, if the order is not changed, the default python implementation has the best performance.

meeteval/io/base.py
boeddeker: I have a preference for the `to_seg_lst_segment`. While it is lengthy,
there are alternatives, that support autocomplete: `SegLstSegment.new(obj)`
The motivation for the existence of `to_seg_lst_segment` is mainly be used internally,
so a longer name is not an issue.

From the `to_dict` I would expect that the key names aren't changed.

thequilo: Okay, I renamed it to `to_seglst_segment` (to match `to_seglst`).

boeddeker (approved):
Except the property keys, the code LGTM.

PR #47, opened 2024-01-05
Add SegLST to load wrapper

thequilo:
Also uses `meeteval.io.load` in the CLI so that the Chime JSON files can be used directly in the CLI:

```bash
meeteval-wer cpwer -r ref.json -h hyp.json
```

### DIFF
--- a/meeteval/io/base.py
+++ b/meeteval/io/base.py
@@ -162,7 +162,7 @@ class Base(BaseABC):
 
     @classmethod
     def merge(cls, *o) -> 'Self':
-        assert all([type(o_) == type(o[0]) for o_ in o]), o
+        assert all([type(o_) == cls for o_ in o]), o
         return cls([line for o_ in o for line in o_.lines])
 
     def _repr_pretty_(self, p, cycle):
--- a/meeteval/io/base.py
+++ b/meeteval/io/base.py
@@ -162,6 +162,7 @@ class Base(BaseABC):
 
     @classmethod
     def merge(cls, *o) -> 'Self':
+        assert all([type(o_) == type(o[0]) for o_ in o]), o
         return cls([line for o_ in o for line in o_.lines])
 
     def _repr_pretty_(self, p, cycle):
--- a/meeteval/io/base.py
+++ b/meeteval/io/base.py
@@ -364,19 +364,3 @@ def _open(f, mode='r'):
         return open(f, mode)
     else:
         raise TypeError(type(f), f)
-
-
-def load(file, parse_float=decimal.Decimal):
-    import meeteval
-    file = Path(file)
-    if file.suffix == '.stm':
-        load_fn = meeteval.io.stm.STM.load
-    elif file.suffix == '.rttm':
-        load_fn = meeteval.io.rttm.RTTM.load
-    elif file.suffix == '.uem':
-        load_fn = meeteval.io.uem.UEM.load
-    elif file.suffix == '.ctm':
-        load_fn = meeteval.io.ctm.CTM.load
-    else:
-        raise ValueError(f'Unknown file type: {file}')
-    return load_fn(file, parse_float=parse_float)
--- a/meeteval/io/base.py
+++ b/meeteval/io/base.py
@@ -160,6 +160,10 @@ class Base(BaseABC):
         # return cls([cls.line_cls.parse(line) for line in s.splitlines() if line.strip()])
         raise NotImplementedError
 
+    @classmethod
+    def merge(cls, *o) -> 'Self':
+        return cls([line for o_ in o for line in o_.lines])
+
     def _repr_pretty_(self, p, cycle):
         name = self.__class__.__name__
         with p.group(len(name) + 1, name + '(', ')'):
--- a/meeteval/io/ctm.py
+++ b/meeteval/io/ctm.py
@@ -1,9 +1,11 @@
 import typing
 import warnings
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Optional
 from meeteval.io.base import Base, BaseLine, BaseABC
 import decimal
+import logging
 
 if typing.TYPE_CHECKING:
     from typing import Self
@@ -15,7 +17,7 @@ __all__ = [
     'CTMGroup',
 ]
 
-
+_warned = False
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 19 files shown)


meeteval/io/seglst.py
```diff
@@ -43,13 +45,66 @@ class SegLST(BaseABC):
...
+        >>> SegLST.parse('[{"words": "a b c", "segment_index": 0, "speaker": 0}]')
+        SegLST(segments=[{'words': 'a b c', 'segment_index': 0, 'speaker': 0}])
+        """
+        import simplejson
+
+        def fix_floats(s):
+            """Convert common float keys to decimal"""
+            for k in ('start_time', 'end_time', 'confidence'):
+                if k in s:
+                    s[k] = decimal.Decimal(s[k])
```
boeddeker, line 74: Why not
```
s[k] = parse_float(s[k])
```
?

Btw. for loading, we don't need simplejson. It is only necessary in dump to avoid a hack of the internals of the json module.

meeteval/wer/__main__.py
```diff
@@ -115,14 +115,12 @@ def _load_hypothesis(hypothesis: 'list[Path]'):
             return CTMGroup.load(hypothesis).to_stm()
         else:
             return CTMGroup.load([hypothesis]).to_stm()
-    elif filename.endswith('.stm'):
-        return STM.load(hypothesis)
     elif filename.startswith('/dev/fd/') or filename.startswith('/proc/self/fd/'):
         # This is a pipe, i.e. python -m ... <(cat ...)
         # For now, assume it is an STM file
         return STM.load(hypothesis)
```
boeddeker, line 121: Add a ToDo to support more formats?

meeteval/wer/__main__.py
```diff
@@ -95,7 +95,7 @@ def _load(path: Path):
 
 def _load_reference(reference: 'Path | list[Path]'):
     """Loads a reference transcription file. Currently only STM supported"""
```
boeddeker, line 97: `Currently only STM supported` <- That is no longer valid

boeddeker:
Could you add some tests to `test_cli`?

meeteval/io/seglst.py
```diff
@@ -43,13 +45,66 @@ class SegLST(BaseABC):
...
+            return s
+        return cls([fix_floats(s) for s in simplejson.loads(s, parse_float=parse_float)])
+
+    def dump(self, file):
+        from meeteval.io.base import _open
+        with _open(file, 'w') as fd:
+            import simplejson
+            simplejson.dump(self.segments, fd, use_decimal=True, indent='    ')
```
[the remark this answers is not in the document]
thequilo: I think 4 was used in the official Chime files. But I'm fine with both, I'll change it to 2

meeteval/wer/__main__.py
thequilo: Where exactly would you add it and which formats would you support in the future?

thequilo:
One open thing is: Should `meeteval.io.load(['1.ctm', '2.ctm'])` return `CTM` or `CTMGroup` with 2 speakers? I'm not sure about that one because we need multiple CTM files in the CLI to be treated as multiple speakers but that would differ from the other formats.

meeteval/io/base.py
```diff
@@ -160,6 +160,10 @@ def parse(cls, s: str, parse_float=decimal.Decimal) -> 'Self':
         # return cls([cls.line_cls.parse(line) for line in s.splitlines() if line.strip()])
         raise NotImplementedError
 
+    @classmethod
+    def merge(cls, *o) -> 'Self':
```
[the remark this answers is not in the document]
thequilo: Good point. After discussion, we decided to keep it as it is.

Reason: It's difficult to make all of the following cases work: `stm1.merge(stm2)`, `STM.merge(stm1, stm2)` and `STM.merge()`. This is possible with descriptors but that's too much effort for this. This issue is present in many other frameworks, as well.

meeteval/io/seglst.py
thequilo: Done

meeteval/wer/__main__.py
thequilo: Done

meeteval/wer/__main__.py
boeddeker: Something like that:
```
# For now, assume it is an STM file
# ToDo: Support more formats that can be clearly distinguished, just from the content.
return STM.load(hypothesis)
```

thequilo: Done

boeddeker:
> One open thing is: Should meeteval.io.load(['1.ctm', '2.ctm']) return CTM or CTMGroup with 2 speakers? I'm not sure about that one because we need multiple CTM files in the CLI to be treated as multiple speakers but that would differ from the other formats.

Hmm. CTM is special, because it doesn't support speakers. But we should support it, because asclite uses it.
Maybe `meeteval.io.load(['1.ctm', '2.ctm'])` returns a `CTMGroup` and prints a Warnung?

`CTM files should only be used, when compatibility with asclite is important. Treat the file stem as speaker name.`

thequilo:
Yes, we could do that. Then, should we move the whole code from `_load_hypothesis` to `meeteval.io.load`, i.e., handling pipes?

boeddeker:
Hmm. The CTM stuff is annoying in a generic load function. The pipe support is only useful, when used from the CLI.
Maybe move the code to load_wrapper and make it privat?

meeteval/io/base.py
```diff
@@ -160,6 +160,11 @@ def parse(cls, s: str, parse_float=decimal.Decimal) -> 'Self':
...
+        assert all([type(o_) == type(o[0]) for o_ in o]), o
```
boeddeker, line 165: Change to `assert all([type(o_) == cls for o_ in o]), o`

meeteval/io/ctm.py
```diff
@@ -105,6 +107,15 @@ class CTM(Base):
...
     @classmethod
     def parse(cls, s: str, parse_float=decimal.Decimal) -> 'Self':
+        global _warned
+        if not _warned:
+            _warned = True
+            logging.warning(
+                'CTM files do not support speaker IDs, so using them is fragile. '
+                'CTM files should only be used when compatibility with other tools (e.g., asclite) is required. '
+                'If compatibility is not important, consider using STM files instead. '
+                'Each CTM file path is treated as a separate speaker.'
```
boeddeker, line 117: Do we use the file path or the file name for the speaker ID?

meeteval/io/seglst.py
```diff
@@ -43,13 +53,66 @@ class SegLST(BaseABC):
...
+        Parses a SegLST from a string.
+
+        >>> SegLST.parse('[{"words": "a b c", "segment_index": 0, "speaker": 0}]')
+        SegLST(segments=[{'words': 'a b c', 'segment_index': 0, 'speaker': 0}])
+        """
+        import simplejson
+
+        def fix_floats(s):
+            """Convert common float keys to decimal"""
+            for k in ('start_time', 'end_time', 'confidence'):
```
boeddeker, line 80: How about the following idea:
Only change keys, that may be used in meeteval?
I would only convert start_time and end_time to float/decimal, because some Code may do strange things with strings.

boeddeker (approved):

meeteval/io/base.py
thequilo: Done

meeteval/io/seglst.py
thequilo: Done

PR #49, opened 2024-01-25
Add sorting options for non-time-constrained WERs

thequilo:


meeteval/io/seglst.py
```diff
@@ -85,9 +87,20 @@ def fix_floats(s):
...
-            fix_floats(s) for s in simplejson.loads(s, parse_float=parse_float)
-        ])
+        loaded = simplejson.loads(s, parse_float=parse_float)
+
+        if not isinstance(loaded, list):
+            raise ValueError(
+                'Invalid JSON format for SegLST: Expected a list of segments, '
+                'but found a dict.'
+            )
+        if loaded and not isinstance(loaded[0], dict):
```
boeddeker, line 97: The first entry in a json is sometimes special. We should also check the last entry (last entry avoids more ifs):

`if loaded and not (isinstance(loaded[0], dict) and isinstance(loaded[-1], dict))`

boeddeker (approved):

meeteval/io/seglst.py
thequilo: Done, also checked for `"session_id"` key.

PR #50, opened 2024-01-25
add api file with a mirror of the shell interface in python

boeddeker:


meeteval/wer/api.py
```diff
@@ -0,0 +1,179 @@
...
+        return match or [pathname]
+
+    def load(paths) -> meeteval.io.SegLST:
+        if isinstance(paths, (str, Path)):
+            paths = [paths]
+        if isinstance(paths, (tuple, list)):
+            paths = [Path(file) for r in paths for file in _glob(str(r))]
+            return meeteval.io.asseglst(
+                meeteval.io.load(paths, format=file_format))
+        elif isinstance(paths, meeteval.io.base.BaseABC):
```
thequilo, line 43: I'm not sure whether this is the correct check. There is a todo on my list for a converter for dataframes, which don't inherit `meeteval.io.base.BaseABC`

meeteval/wer/api.py
```diff
@@ -0,0 +1,179 @@
...
+        hypothesis_sort: 'bool | str' = False,
+        file_format=None,
+) -> 'tuple[meeteval.io.SegLST, meeteval.io.SegLST]':
+    """Load and validate reference and hypothesis texts.
+
+    Validation checks that reference and hypothesis have the same example IDs.
+    """
+
+    # Normalize and glob (for backwards compatibility) the path input
+    def _glob(pathname):
```
thequilo, line 31: Do we keep the globbing? I think that this can cause weird behavior. Especially because `_save_results` does not perform globbing currently

meeteval/wer/api.py
```diff
@@ -0,0 +1,179 @@
...
+        match = list(glob.glob(pathname))
+        # Forward pathname if not matched to get the correct error message
+        return match or [pathname]
+
+    def load(paths) -> meeteval.io.SegLST:
```
thequilo, line 36: Move this to the top level?

tests/test_py.py
thequilo, line 1: I think the name `test_api.py` or `test_python_api.py` would be clearer

tests/test_py.py
```diff
@@ -0,0 +1,104 @@
...
+    avg = combine_error_rates(details)
+    check_result(avg, 88, 92)
+
+    details = tcpwer(meeteval.io.load(ref), meeteval.io.load(hyp))
+    avg = combine_error_rates(details)
+    check_result(avg, 88, 92)
+
+    details = tcpwer(meeteval.io.load(ref).to_seglst(), meeteval.io.load(hyp).to_seglst())
+    avg = combine_error_rates(details)
+    check_result(avg, 88, 92)
```
thequilo, line 58: Test other options, i.e., collar and pseudo-word-level strategies

meeteval/wer/api.py
boeddeker: I will change it to try `meeteval.io.asseglst` and raise an TypeError, if that function raises an Excpetion.

meeteval/wer/api.py
boeddeker: I thought it is not difficult to keep it, so keeping means to not introduce a breaking change.
In `_save_results` it shouldn't be an issue. It can still create a reasonable name.

meeteval/wer/api.py
boeddeker: Any suggestion for a name in top level? Maybe something like `_maybe_load`?

tests/test_py.py
boeddeker: Done

tests/test_py.py
boeddeker: Done

meeteval/wer/api.py
thequilo: Okay, the name just might be unexpected and contain `*`

meeteval/wer/api.py
thequilo: I'm fine with that!

boeddeker: Done

meeteval/wer/api.py
```diff
@@ -0,0 +1,181 @@
...
+        return meeteval.io.asseglst(
+            meeteval.io.load(paths, format=file_format))
+    else:
+        try:
+            return meeteval.io.asseglst(paths)
+        except Exception as e:
+            raise TypeError(type(paths), paths) from e
+
+
+def _load_texts(
```
thequilo, line 38: Are there any operations that might span across reference and hypothesis? Currently there's not, and then we could simplify to the signature `_load_text(paths, regex, sort, file_format)` and call it one time for the reference and one time for the hypothesis

thequilo: Is `_load_text` even a good name anymore? It now supports much more than text, e.g., RTTM. `_load` could be enough, although it clashes with `_maybe_load`

thequilo (approved):

PR #51, opened 2024-01-25
Fix zero time interval in example

boeddeker:


thequilo:
This was intended as a corner case. Does it break something?

boeddeker:
The visualisation is suboptimal for our toy example. It is difficult to see an utterance at 0. Was the corner case zero interval at zero or just zero interval?

thequilo:
The corner case was just zero duration for any segment. You can also add an empty segment with zero duration

boeddeker:
Ok, I moved the zero duration segment now to 1-1. There it is better recognizable. I also added an empty segment.

thequilo (approved):

PR #52, opened 2024-01-26
Add first version of time-constrained ORC-WER (tcORC-WER)

thequilo:
This PR adds code to compute the tcORC-WER.

Example:

```shell
python -m meeteval.wer tcorcwer -h hyp.stm -r ref.stm --collar 5
```

The current version is tested on Libri-CSS with a system that produces 8 streams. It finished computation within 10 minutes and used less than 2GB of RAM (which is a huge improvement over ORC-WER!). These requirements should drop further when the number of streams is smaller.

The code is not fully optimized yet and contains many TODOs. I'll work on some of these TODOs and update the PR during the next few days.

I'll merge main back into this PR once #50 is merged.

### DIFF
--- a/meeteval/wer/wer/time_constrained_orc.py
+++ b/meeteval/wer/wer/time_constrained_orc.py
@@ -1,0 +1,122 @@
+import meeteval
+from meeteval.wer import ErrorRate
+from meeteval.wer.wer.orc import OrcErrorRate
+from meeteval.wer.wer.utils import check_single_filename
+
+
+def time_constrained_orc_wer(
+        reference,
+        hypothesis,
+        collar=0,
+        reference_pseudo_word_level_timing='character_based',
+        hypothesis_pseudo_word_level_timing='character_based_points',
+        hypothesis_sort='segment',
+):
+    """
+    The time-constrained version of the ORC-WER (tcORC-WER).
+
+    Special cases where the reference or hypothesis is empty
+    >>> time_constrained_orc_wer([], [])
+    OrcErrorRate(errors=0, length=0, insertions=0, deletions=0, substitutions=0, assignment=())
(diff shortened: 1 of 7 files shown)


meeteval/wer/wer/time_constrained_orc.py
```diff
@@ -0,0 +1,145 @@
...
+    check_single_filename(reference, hypothesis)
+
+    # Remove empty segments
+    reference = reference.filter(lambda s: s['words'] != '')
+    hypothesis = hypothesis.filter(lambda s: s['words'] != '')
+
+    # Add a segment index to the reference so that we can later find words that
+    # come from the same segment
+    for i, s in enumerate(reference):
+        s['segment_index'] = i
```
boeddeker, line 57: Hmm. Should the assignment be before the filter?

tests/test_time_constrained_orc_matching.py
```diff
@@ -0,0 +1,96 @@
...
+    # assert orc.deletions == tcorc.deletions
+    # assert orc.substitutions == tcorc.substitutions
+
+
+@given(
+    seglst(max_speakers=1, min_segments=1),
+    seglst(max_speakers=2, min_segments=1),
+)
+def test_orc_bound_by_tcorc(reference, hypothesis):
+    """Tests that tcORC-WER is never lower than ORC-WER"""
```
boeddeker, line 71: Add also the tcp upper bound to the tests?

ORC-WER <= tcORC-WER <= tcpWER

meeteval/wer/wer/time_constrained_orc.py
thequilo: It shouldn't make a difference because the segment index is here just used for grouping and not for sorting. But it could be easier to understand what's happening when moved before the filter operation

boeddeker (approved):

PR #53, opened 2024-01-29
viz: Add support for recording audio file(s)

boeddeker:
Bildschirmfoto vom 2024-01-29 13-54-32

### DIFF
--- a/meeteval/viz/__main__.py
+++ b/meeteval/viz/__main__.py
@@ -3,33 +3,40 @@ import collections
 
 import meeteval
 from meeteval.viz.visualize import AlignmentVisualization
+from meeteval.wer.api import _load_texts
 import tqdm
+
 
 def create_viz_folder(
         reference,
         hypothesiss,
         out,
         alignment='tcp',
+        regex=None,
 ):
     out = Path(out)
-    if isinstance(reference, (str, Path)):
-        reference = meeteval.io.load(reference).to_seglst()
-
-    reference = reference.groupby('session_id')
     out.mkdir(parents=True, exist_ok=True)
 
     avs = {}
     for i, hypothesis in tqdm.tqdm(hypothesiss.items()):
-        if isinstance(hypothesis, (str, Path)):
-            hypothesis = meeteval.io.load(hypothesis).to_seglst()
 
-        hypothesis = hypothesis.groupby('session_id')
+        r, h = _load_texts(
+            reference, hypothesis, regex=regex,
+            reference_sort='segment',
+            hypothesis_sort='segment',
+        )
 
-        session_ids = set(reference.keys()) & set(hypothesis.keys())
+        r = r.groupby('session_id')
+        h = h.groupby('session_id')
+
+        session_ids = set(r.keys()) & set(h.keys())
+        xor = set(r.keys()) ^ set(h.keys())
+        if xor:
+            print(f'Ignore {xor}, because they are not available in reference and hypothesis.')
 
         for session_id in session_ids:
-            av = AlignmentVisualization(reference[session_id],
-                                        hypothesis[session_id],
+            av = AlignmentVisualization(r[session_id],
+                                        h[session_id],
                                         alignment=alignment)
             av.dump(out / f'{session_id}_{i}.html')
             avs.setdefault(i, {})[session_id] = av
@@ -89,13 +96,7 @@ def create_viz_folder(
         with tag('body'):
             with tag('table', klass='tablesorter', id='myTable'):
                 with tag('thead'), tag('tr'):
-
-                    # meeteval.wer.combine_error_rates([
-                    #         meeteval.wer.ErrorRate.from_dict(av.data['info']['wer'] for av in avs.values())
-                    #
-                    #     ])
                     def get_wer(v):
-                        print(list(v.values())[0].data['info']['wer']['hypothesis'])
                         error_rate = meeteval.wer.combine_error_rates(*[
                             meeteval.wer.ErrorRate.from_dict(av.data['info']['wer']['hypothesis'])
                            for av in v.values()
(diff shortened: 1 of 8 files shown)


meeteval/viz/file_server.py
```diff
@@ -1,35 +1,56 @@
 """
-Deliver normalized audio files.
+A file server, that exposes wav files from the filesystem:
```
thequilo, line 2: Remove comma

meeteval/viz/file_server.py
```diff
@@ -1,35 +1,56 @@
...
+ - Only delivers wav files
+ - Normalizes the audio files (Useful for far files, that is typically low volumn)
```
thequilo, line 4: Normalizes the audio files (Useful for far-field recordings that typically have a low volume)

meeteval/viz/visualize.py
```diff
@@ -352,7 +354,12 @@ def __init__(
             highlight_regex=None,
             alignment_transform=None,
             markers=None,
+            recording_file='',
```
thequilo, line 357: Set the default to None?

tests/test_python_api.py
```diff
@@ -67,10 +67,10 @@ def test_tcpwer():
     # There is no motivation for the selected combinations.
     # Hint: We recommend to use the defaults.
     for h_pwt, r_pwt, errors in [
-        ['character_based_points', 'character_based', 40],
+        ['character_based_points', 'character_based', 42],
```
thequilo, line 70: I also fixed this in the tcORC-WER PR. I hope that this doesn't cause merge conflicts

meeteval/viz/visualize.js
```diff
@@ -1489,6 +1491,45 @@ class CanvasPlot {
...
 
+    function drawRecordingAudioButton(container, rangeSelector, name, recording_file) {
+        let update_audio = function (){
+            // Note: Deleting and adding an audio with the same content doesn't
+            //       trigger a load. Hence, no optimization necessary.
+            audio_div.selectAll("*").remove();
+            let lower = rangeSelector.lower_input.node().value;
+            let upper = rangeSelector.upper_input.node().value;
+            let value = "[" + lower + ":" + upper + "]";
+            let path = server_path.node().value + "/" + file_path.node().value + "::" + value;
```
thequilo, line 1502: This fails when the file path already contains a slicing operation, doesn't it? I think it would be better to use URL arguments: https://docs.aiohttp.org/en/stable/client_quickstart.html#passing-parameters-in-urls

```
server_url/file.wav?start=0&stop=16000
```

meeteval/viz/visualize.js
```diff
@@ -1489,6 +1491,45 @@ class CanvasPlot {
...
+            }
+        };
+
+        let pill = container.append("div").classed("pill", true);
+        let audio_tooltip = addTooltip(pill);
+
+        pill.append("span").text((name.trim() !== "") ? "🔊 " + name : "🔊");
+
+        let input = audio_tooltip.append("div").style("display", "flex");
+        let server_path = input.append("input").attr("type", "text").attr("placeholder", "e.g. http://localhost:7777").property("value", "http://localhost:7777");
```
thequilo, line 1520: This server_path should also affect the segment audios. I just don't know how to pass it through

meeteval/viz/visualize.js
```diff
@@ -1489,6 +1491,45 @@ class CanvasPlot {
...
+        let file_path = input.append("input").attr("type", "text").style("width", "30em").attr("placeholder", "e.g. /path/to/file.wav").property("value", recording_file);
+
+        let audio_div = audio_tooltip.append("div").style("display", "flex");
+
+        input.append("button").text("Load").on("click", () => update_audio());
+        server_path.on("keydown", (event) => {if (event.key === "Enter") {update_audio();}});
+        file_path.on("keydown", (event) => {if (event.key === "Enter") {update_audio();}});
+        pill.on("click", () => {update_audio(); audio_div.select("audio").node().play()});
+
```
thequilo, line 1529: Changing the selection does not update the audio. This might turn out to be annoying. But it's not that easy to get the update in here. You would have to get the details plot, add a callback to onScroll, setup a timer to throttle server requests (update after a certain delay)...

meeteval/viz/file_server.py
boeddeker: done

meeteval/viz/file_server.py
boeddeker: done

meeteval/viz/visualize.py
boeddeker: done

tests/test_python_api.py
boeddeker: Hmm. We should try to remove the red badge in the readme. I was not sure, when you plan to merge tcorc.

meeteval/viz/visualize.js
boeddeker: In General, it would be good to have a settings menu, that does some caching (Across files).
But I don't know how to do that or if it is possible (We don't want server interactions).

meeteval/viz/visualize.js
boeddeker: I initially had the same idea. But it turned out, that I always clicked on the "pill".

So I made the pill to a button, that triggers an update and the play.

Alternatively, the audio could be updated, when the tooltip appears (the browser does some caching, so it will be relative cheap.).

tests/test_python_api.py
thequilo: I could merge it now and open another PR for any further changes to the tcORC-WER

meeteval/viz/visualize.js
thequilo: Updating the audio when the tooltip appears sounds like a good idea

meeteval/viz/visualize.js
boeddeker: Yes, it would fail. I changed it to arguments.
The server accepts now both, the slicing and the arguments.
The js code uses the arguments, but some paths may use the slicing, hence I kept the slicing code.
If both are provided, the code will fail.

Question:
Should both be support double slicing?

e.g. CSS pipeline on UEM data and the user wants to listen to enhanced data and observation.
I am not sure, it that is too special.

meeteval/viz/visualize.js
boeddeker: Hmm. I am not sure, how useful it is, when it is only kept for one URL.

meeteval/viz/visualize.js
boeddeker: Now, where I think about this, it might be better to load the audio only, when it is likely to be used. Just moving the mouse over the pill should not trigger a load.

meeteval/viz/visualize.js
thequilo: I guess it's not super useful, but it's the only way to persist data. Browsers are not made for local files...

meeteval/viz/visualize.js
thequilo: What happens if a user hovers over the pill and starts the audio using the audio controls in the tooltip? The user would then get the old audio. I think starting the audio doesn't trigger a reload

meeteval/viz/visualize.js
boeddeker: Ok, I added the support for both arguments. I kept negative slices for the future.

meeteval/viz/visualize.js
boeddeker: I fixed this now:
When moving the mouse over the pill, it will be checked, whether the content is valid. If not, it will be deleted.
In this way, the number of triggered audio loads is as small as possible.

meeteval/viz/__main__.py
```diff
@@ -0,0 +1,211 @@
...
+        with tag('html'):
+            with tag('style'):
+                doc.asis("""html, body {height: 100%; margin: 0;}""")
+                doc.asis(
+                    """.full-height {height: 100%; background: yellow;}""")
+            with tag('body',
+                     style="width: 100%; height: 100%; display: flex;"):
+                for i, av in v.items():
+                    with tag('div',
+                             style=f"width: {100 // len(v)}%; height: 100%;"):
```
thequilo, line 72: You can use the flex layout to have a varying number of columns without explicit computations. Example:
```html
<html>
<style>
    .grid {
        height: 100%;
        display: flex;
    }
    .column {
        flex-grow: 1;
    }
    div {
        border: 1px solid black;
    }
</style>
<body>
<div class="grid">
    <div class="column">Column 1</div>
    <div class="column">Column 2</div>
    <div class="column">Column 3</div>
</div>
</body>
</html>
```

meeteval/viz/file_server.py
```diff
@@ -1,35 +1,94 @@
 """
-Deliver normalized audio files.
+A file server that exposes wav files from the filesystem:
+ - Only delivers wav files
+    - Support slices (int -> samples, float -> seconds)
+        - server_url/file.wav?start=0&stop=16000
+        - server_url/file.wav::[0:16000]
+ - Normalizes the audio files (Useful for far-field recordings that typically have a low volume)
+ - Supports slicing
```
thequilo, line 8: Slicing support is repeated

meeteval/viz/visualize.js
thequilo: Okay, sounds good!

meeteval/viz/__main__.py
```diff
@@ -0,0 +1,211 @@
...
+    avs_T = dict(avs_T)
+
+    for session_id, v in avs_T.items():
+        doc, tag, text = Doc().tagtext()
+        doc.asis('<!DOCTYPE html>')
+        with tag('html'):
+            with tag('style'):
+                doc.asis("""html, body {height: 100%; margin: 0;}""")
+                doc.asis(
+                    """.full-height {height: 100%; background: yellow;}""")
```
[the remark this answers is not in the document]
boeddeker: No, removed

meeteval/viz/__main__.py
boeddeker: Changed.

meeteval/viz/file_server.py
boeddeker: removed

thequilo (approved):

PR #54, opened 2024-01-30
Add UEM support to WER metrics

thequilo:


boeddeker:
@popcornell What do you think, how the uem files should work?

Especially, what should be done in the corner cases.

I checked your code from the last challenge and there you used pyannote and if I get it right,
they use an "intersection" based cropping, i.e. the segment at the boundaries are replaced by the intersection of
the uem region and the segment. This has an effect on the word positions.

At the moment, we only remove segments that are fully outside of the scoring regions.

popcornell:
tcpWER you derive estimated word positions no ? 
So theoretically you can exclude words that are outside (or on the boundary) and keep ones full inside. 

Another option is maybe weight the current utterance WER statistics by the amount of intersection. 
E.g. segment is half inside the weight is 0.5 for the words in that segment. 

Both are probably complicated to implement ? And the impact is usually very limited. 

> we only remove segments that are fully outside of the scoring regions.
Maybe you can even remove segments that intersect the UEM altogether, a more conservative approach. 
You usually trim based on UEM and thus if you don't remove it the model will get the part outside the UEM wrong for sure. 
Since in long form ASR number of utterances is quite high discarding one or two won't make much difference and is better to be conservative IMO.

popcornell:
> Actually, we thought we are conservative. A user can always manually delete segments, that partially overlap with the scoring region. The words outside of the scoring region could still match with a word inside the scoring region, e.g. some SOT systems tend to estimate delayed time stamps, that could exceed the scoring region, but are still within the collar.

There could be also another speaker outside the scoring region and that would introduce errors. 
But it is true that the user should be careful and handle that knowing the UEM boundaries. 
He may choose to keep or discard them. 
Not scoring the hypotheses that partially overlap also is "hackable" IMO (depends on WER metric but I can devise some loopholes for some) and should not be done. Your approach is most definitely the correct one

boeddeker:
Yes, independent of what we do, it might be wrong for one or another system/user.
A motivation for keeping segments, is that in case of an error, it is very likely that WER will be larger. So the user has to take care to fix it to get a smaller number.

Dropping may yield a smaller number or larger number.

popcornell:
> Dropping may yield a smaller number or larger number.

yes which may be hackable, I may choose to purposedely make my segments as large as possible towards the boundaries so they are not scored. This will not work with tcpWER but may work with some other metrics.

thequilo:
From the discussion, I take away that it's unclear how to handle the edge cases and I'll keep the current processing. If we later decide that another option is better, we can add an option to switch the strategy for applying uem information.

PR #57, opened 2024-02-02
Fix self-overlap computation for tcORC-WER

thequilo:


boeddeker:
Could you add a test, that the overlap is zero for the toy examples?

boeddeker (approved):

PR #60, opened 2024-02-02
Update rttm.py

jkienegger:
RTTMLine can read lines missing final entries, making it compatible to paderbox rttm-tools

### DIFF
--- a/meeteval/io/rttm.py
+++ b/meeteval/io/rttm.py
@@ -56,10 +56,15 @@ class RTTMLine(BaseLine):
         """
         >>> RTTMLine.parse('SPEAKER CMU_20020319-1400_d01_NONE 1 130.430000 2.350 <NA> <NA> juliet <NA> <NA>')
         RTTMLine(type='SPEAKER', filename='CMU_20020319-1400_d01_NONE', channel='1', begin_time=Decimal('130.430000'), duration=Decimal('2.350'), orthography='<NA>', speaker_type='<NA>', speaker_id='juliet', confidence='<NA>', signal_look_ahead_time='<NA>')
+        >>> RTTMLine.parse('SPEAKER CMU_20020319-1400_d01_NONE 1 130.430000 2.350 <NA> <NA> juliet')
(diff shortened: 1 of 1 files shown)


meeteval/io/rttm.py
```diff
@@ -56,10 +56,15 @@ def parse(cls, line: str, parse_float=decimal.Decimal) -> 'RTTMLine':
         """
         >>> RTTMLine.parse('SPEAKER CMU_20020319-1400_d01_NONE 1 130.430000 2.350 <NA> <NA> juliet <NA> <NA>')
         RTTMLine(type='SPEAKER', filename='CMU_20020319-1400_d01_NONE', channel='1', begin_time=Decimal('130.430000'), duration=Decimal('2.350'), orthography='<NA>', speaker_type='<NA>', speaker_id='juliet', confidence='<NA>', signal_look_ahead_time='<NA>')
+        >>> RTTMLine.parse('SPEAKER CMU_20020319-1400_d01_NONE 1 130.430000 2.350 <NA> <NA> juliet')
+        RTTMLine(type='SPEAKER', filename='CMU_20020319-1400_d01_NONE', channel='1', begin_time=Decimal('130.430000'), duration=Decimal('2.350'), orthography='<NA>', speaker_type='<NA>', speaker_id='juliet', confidence='<NA>', signal_look_ahead_time='<NA>')
```
thequilo, line 60: Can you add documentation where you found that the trailing fields are optional? I looked in dscore (https://github.com/nryant/dscore?tab=readme-ov-file#rttm) and the NIST docs (https://web.archive.org/web/20170119114252/http://www.itl.nist.gov/iad/mig/tests/rt/2009/docs/rt09-meeting-eval-plan-v2.pdf) and both don't mention that fields are optional. The NIST docs say that you should set the fields to `<NA>` when a system doesn't provide the information

boeddeker: In https://catalog.ldc.upenn.edu/docs/LDC2004T12/RTTM-format-v13.pdf is the RTTM definition with only 9 fields.

boeddeker: Additionally, kaldi also uses 9 fields: https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/steps/segmentation/convert_utt2spk_and_segments_to_rttm.py

thequilo (approved):

meeteval/io/rttm.py
thequilo: All right, these formats are a total mess

boeddeker: Not totally. At least they are specified and are some kind of standard.

PR #63, opened 2024-02-06
Fix apply_multi_file for empty reference and missing hypothesis keys

thequilo:


meeteval/io/seglst.py
```diff
@@ -576,47 +584,71 @@ def apply_multi_file(
         h_minus_r = list(set(hypothesis.keys()) - set(reference.keys()))
         r_minus_h = list(set(reference.keys()) - set(hypothesis.keys()))
 
-        ratio = len(r_minus_h) / len(reference.keys())
+        if len(reference.keys()) == 0:
+            logging.warning(
+                f'The reference is empty, but found keys '
+                f'{list(hypothesis.keys())} in the hypothesis'
+            )
+            ratio = 0
```
boeddeker, line 592: Hmm. What is the motivation to allow an empty reference?
I will run in the `ratio <= allowed_empty_examples_ratio` if statement and only print a warning.

Shouldn't be the check here: `if len(reference.keys()) == 0 and len(hypothesis.keys()) == 0`? But I am not sure, if it is possible to hit this case.

meeteval/io/seglst.py
```diff
@@ -576,47 +584,71 @@ def apply_multi_file(
...
-                f'Continue with the assumption, that the system predicted '
-                f'silence for the missing recordings.',
+                f'Missing {ratio * 100:.3f} % = '
+                f'{len(r_minus_h)}/{len(reference.keys())} of recordings in '
+                f'hypothesis. An exception will be raised when this number '
+                f'exceeds {allowed_empty_examples_ratio * 100:3} %.\n'
+                f'Please check if your system ignored recordings or '
+                f'predicted silence (no transcripts) for some recordings. '
+                f'It is recommended to output an empty transcript if silence '
+                f'was recognized.\n'
```
boeddeker, line 632: `... was recognized, to be able to reliably detect an error.` Adding something like that? Or is that obvious?

meeteval/io/seglst.py
```diff
@@ -576,47 +584,71 @@ def apply_multi_file(
...
-                f'hypothesis - reference: e.g. {h_minus_r[:5]} '
-                f'(Total: {len(h_minus_r)} of {len(hypothesis)})\n'
-                f'reference - hypothesis: e.g. {r_minus_h[:5]} '
-                f'(Total: {len(r_minus_h)} of {len(reference)})'
+                f'Missing {ratio * 100:.3f} % = '
+                f'{len(r_minus_h)}/{len(reference.keys())} of recordings in '
+                f'hypothesis. This exceeds the threshold of '
+                f'{allowed_empty_examples_ratio * 100:3} %.\n'
+                f'Please check if your system ignored recordings or '
+                f'predicted silence (no transcripts) for some recordings. '
```
boeddeker, line 644: Is empty transcript fine?

`Please check if your system ignored recordings or predicted nothing (no transcripts) for some recordings`

My interpretation: `silence` -> `empty transcript`

tests/test_python_api.py
```diff
@@ -191,3 +192,49 @@ def test_md_eval_22():
...
+
+    # Check missing hyp amount above ratio
+    m = Mock()
+    with pytest.raises(RuntimeError):
+        apply_multi_file(m, ref, hyp.filter(lambda x: x.filename == 'recordingA'), allowed_empty_examples_ratio=0)
+    assert m.session_ids == []
+
+    # Check empty reference
+    m = Mock()
+    apply_multi_file(m, meeteval.io.SegLST([]), hyp)
```
boeddeker, line 234: Should this fail? Doesn't it hide an error?

meeteval/io/seglst.py
thequilo: Hmm, right, I just wanted to fix this case because it failed in random tests. But it could be better to let this case fail. On the other hand, it could be annoying if it fails for an empty reference. In most cases it'll be a usage error, but there are rare cases where you want an empty reference. 

An alternative would be to disallow empty references and enforce that the user filters the inputs correctly, e.g., with the `--regex` argument

thequilo: I meant disallowing missing keys in the reference, which is currently interpreted as intended by the user and dropped in the hypothesis

meeteval/io/seglst.py
thequilo: Done

meeteval/io/seglst.py
thequilo: I removed the second part of the sentence. I think that makes it easier to understand

tests/test_python_api.py
```diff
@@ -191,3 +192,57 @@ def test_md_eval_22():
...
+    with pytest.raises(RuntimeError):
+        apply_multi_file(m, ref, hyp.filter(lambda x: x.filename == 'recordingA'), allowed_empty_examples_ratio=0)
+    assert m.session_ids == []
+
+    # Check empty reference
+    with pytest.raises(RuntimeError):
+        m = Mock()
+        apply_multi_file(m, meeteval.io.SegLST([]), hyp)
+    m = Mock()
+    apply_multi_file(m, meeteval.io.SegLST([]), hyp, partial=True)
```
boeddeker, line 240: Why should this code be executable? I don't see a motivation, where this is useful. It will only hide an error.
Additionally, in the cli, the code gets problems, because the average cannot be computed.

boeddeker (approved):

Issue #62, opened 2024-02-06
Does mdeval computes DER taking overlapped speech into account?


AntoineBlanot:
Tools like [pyannote](https://github.com/pyannote/pyannote-metrics) allows us to choose the collar and if we want to compute DER on overlapped speech regions or not.

With `mdeval`, we can specify the collar but it seems like there is no option for including overlapped speech in the metric or not.
Does that mean that by default it computes over overlapped regions? Or are they excluded for the calculations?

Thank you for your answer !

thequilo:
Hi @AntoineBlanot! I had to check that first. `md-eval-22.pl` has an option `-o` 

> -o to include overlapping speech in MD evaluation.  With this option, separate recognition passes are made for each reference speaker.

This option is, however, currently not set by our wrapper.

Even if the option is not set, I found the following things:

- md_eval ignores self-overlap and treats these regions as if the speaker was active continuously. It does warn about self-overlap though.
```python
meeteval.der.md_eval_22(
    meeteval.io.asseglst([
        {'speaker': 'A', 'start_time': 0, 'end_time': 7, 'session_id': 'X', 'words': ''},
        {'speaker': 'A', 'start_time': 4, 'end_time': 10, 'session_id': 'X', 'words': ''},
    ]),
    meeteval.io.asseglst([{'speaker': 'A', 'start_time': 0, 'end_time': 10, 'session_id': 'X', 'words': ''}]),
)
# WARNING:  speaker A speaking more than once at time 4
# WARNING:  speaker A speaking more than once at time 4
# WARNING:  speaker A speaking more than once at time 4
# WARNING:  speaker A speaking more than once at time 4
# {'X': DiaErrorRate(error_rate=Decimal('0.00'), scored_speaker_time=Decimal('10.000000'), missed_speaker_time=Decimal('0.000000'), falarm_speaker_time=Decimal('0.000000'), speaker_error_time=Decimal('0.000000'))}
```
- md_eval seems to compute DER for overlapping regions, even if `-o` is not set (`scored_spaker_time` is 20).
```python
meeteval.der.md_eval_22(
    meeteval.io.asseglst([
        {'speaker': 'A', 'start_time': 0, 'end_time': 10, 'session_id': 'X', 'words': ''},
        {'speaker': 'B', 'start_time': 0, 'end_time': 10, 'session_id': 'X', 'words': ''},
    ]),
    meeteval.io.asseglst([
[...]

PR #64, opened 2024-02-15
Add argument to only evaluate single-speaker regions

thequilo:
Adresses #62. The name is still up for discussion.

### DIFF
--- a/meeteval/der/__main__.py
+++ b/meeteval/der/__main__.py
@@ -7,6 +7,7 @@ def md_eval_22(
         average_out='{parent}/{stem}_md_eval_22.json',
         per_reco_out='{parent}/{stem}_md_eval_22_per_reco.json',
         collar=0,
+        exclude_overlap=False,
         regex=None,
         uem=None,
 ):
@@ -16,6 +17,7 @@ def md_eval_22(
         hypothesis,
         collar=collar,
         regex=regex,
+        exclude_overlap=exclude_overlap,
         uem=uem,
     )
     _save_results(results, hypothesis, per_reco_out, average_out)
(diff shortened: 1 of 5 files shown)


meeteval/der/__main__.py
```diff
@@ -24,7 +26,20 @@ def md_eval_22(
...
     from meeteval.wer.__main__ import CLI
 
-    cli = CLI()
+    class DerCLI(CLI):
+        def add_argument(self, command_parser, name, p):
+            if name == 'exclude_overlap':
+                command_parser.add_argument(
+                    '-1', '--exclude-overlap',
+                    action='store_true',
+                    help='Limits scoring to single-speaker regions. '
```
boeddeker, line 35: @TCord Could you add a sentence, where the user should be careful? Or was that only relevant, if you want to calculate the overlap only DER?

boeddeker:
I vote for`--regions [all|single|overlap|nonoverlap]` instead of adding a new option that is not used in any existing framework.
It is already used in spyder and works for md-eval, dscore, pyannote and spyder.

thequilo:
I also like `--regions`, but md-eval only supports `--regions [all|nooverlap]`. Also, should we name it `--score-regions`?

boeddeker:
> I also like --regions, .... Also, should we name it --score-regions?

Hmm. My motivation was to use the same name as others.

> but md-eval only supports `--regions [all|nooverlap]`

That shouldn't be a problem.

thequilo:
> Hmm. My motivation was to use the same name as others.

`--regions` is not as self-explanatory, `--score-regions` gives a much better hint at what this "region" means. But since all names we found so far have problems, I'm pretty sure that anything we'll come up with will also have its flaws, so I'll stick with `--regions`.

boeddeker (approved):

PR #66, opened 2024-02-15
Viz: Make the visualization's height responsive

thequilo:
Adjust the details plot height such that the visualization fits the screen size.

Uses an approximation in the notebook: Assume that the notebook UI uses ~20% of the available window height and scale the visualization to `80vh`.


thequilo:
Related to this, I also changed:

- Made Minimaps resizable by dragging the x-axis. This is relevant for the responsive height feature because the minimap can now fill up the whole space, leaving no space for the details plot. The user can now freely choose the height of the minimaps
- Improve dragging in the details plot. The mouse movements have to be tracked globally because the mouse can briefly move over the details plot when dragging the lowest minimap's x-axis for resizing. This would cause a resize and drag happen at the same time, which we don't want. The mouse can now leave the details plot while dragging and the drag continues, which is a much better user experience. Also deactivated text selection while dragging
- Fix scrolling and zooming and implement the minimap resize feature for touch devices.

PR #70, opened 2024-03-28
Centralize viewArea state in Visualization

thequilo:
The viewAreas are now managed in a central location instead of in the local minimap objects. The central state is a single source of truth and eliminates much of the communication between plot objects. This has a few benefits:
- much shorter implementations for the minimaps (because the communication code is gone)
- reasoning about state changes is easier because everything is handled in a central location
- throttling is synchronized between plots and applied to both word filtering and plotting
- It is easier to determine which plots have to be updated, which results in significant speedups for viewing long recordings 

Additionally, functionality was added to synchronize plots in a side-by-side view.

### DIFF
--- a/meeteval/viz/visualize.js
+++ b/meeteval/viz/visualize.js
@@ -293,7 +293,13 @@ function alignment_visualization(
         const match = /^(?<start>[+-]?([0-9]*[.])?[0-9]+)\s*-\s*(?<end>[+-]?([0-9]*[.])?[0-9]+)$/.exec(selection);
 
         if (match) {
-            return [parseFloat(match.groups.start), parseFloat(match.groups.end)];
+            const start = parseFloat(match.groups.start);
+            const end = parseFloat(match.groups.end);
+
+            // Check for invalid ranges. Return null if invalid.
+            if (isNaN(start) || isNaN(end) || start > end) return null;
+
+            return [start, end];
         }
     }
 
@@ -323,7 +329,11 @@ function alignment_visualization(
         if (finalViewArea === null) {
             if (urlParams.has('selection')) {
                 console.log("Setting selection from URL", urlParams.get('selection'));
-                finalViewArea = parseSelection(urlParams.get('selection')) ?? domain;
+                finalViewArea = parseSelection(urlParams.get('selection'));
+                if (!finalViewArea) {
+                    console.log("Invalid selection in URL. Using full domain.");
+                    finalViewArea = domain;
+                }
             } else {
                 finalViewArea = domain;
             }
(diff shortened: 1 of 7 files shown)


meeteval/viz/visualize.js
```diff
@@ -23,6 +23,169 @@ var colormaps = {
...
+/**
+ * Checks whether two ranges `a` and `b` are similar by relative tolerance `tolerance`.
+ */
+function similar_range(a, b, tolerance=0.00001){
+    // Note: a == [0, 1] will always return false, because arrays are compared by reference, not values
+    if (JSON.stringify(a) === JSON.stringify(b)) {
+        return true
+    }
+
+    if (!a || a.some(isNaN) || a.some(isNaN))
```
boeddeker, line 107: Why do you test only a?

boeddeker: a.some(isNaN) is repeated

meeteval/viz/visualize.js
```diff
@@ -23,6 +23,169 @@ var colormaps = {
...
+    if (viewArea) {
+        viewAreas[anchor] = viewArea;
+        dirty[anchor] = true;
+    }
+
+    // Update children of anchor (j > i)
+    for (let j = anchor + 1; j < viewAreas.length; j++) {
+        let v = viewAreas[j];
+        const parent = viewAreas[j - 1];
+        if (parent[1] - parent[0] < v[1] - v[0] || similar_range(oldViewAreas[j - 1], v)) {
```
boeddeker, line 144: Should it be marked as dirty, when the values ares the same? A draw wouldn't be necessary.

meeteval/viz/visualize.js
```diff
@@ -23,6 +23,169 @@ var colormaps = {
...
+ * Ensures that every entry (index i) lies within its parent (index i - 1).
+ *
+ * Keeps the viewArea at `anchor` fixed.
+ *
+ * Returns an array of booleans indicating which viewAreas were changed.
+ */
+function adjustViewAreas(viewAreas, anchor, viewArea) {
+    const dirty = viewAreas.map(() => false);
+
+    const oldViewAreas = [...viewAreas];
```
boeddeker, line 133: I am not sure, if it works as intended. You do inplace changes and keep only the old value if the child is larger than the parent.
If this is the  indented behavior, it should be documented.

meeteval/viz/visualize.js
```diff
@@ -23,6 +23,169 @@ var colormaps = {
...
+    for (let j = anchor - 1; j > 0; j--) {
+        let v = viewAreas[j];
+        const child = viewAreas[j + 1];
+        if (child[1] - child[0] > v[1] - v[0]) {
+            v = child;
+            dirty[j] = true;
+        } else if (v[0] > child[0]) {
+            const diff = child[0] - v[0];
+            viewAreas[j][0] += diff;
+            viewAreas[j][1] += diff;
```
boeddeker, line 175: change v?

meeteval/viz/visualize.js
thequilo: Done

meeteval/viz/visualize.js
thequilo: Solved, now only marked as dirty when `!similar_rate(viewAreas[j], v)`

PR #71, opened 2024-04-02
viz: move file_server path to settings and support relative paths

boeddeker:


meeteval/viz/visualize.js
```diff
@@ -632,6 +633,23 @@ class CanvasPlot {
...
+
+
+        // Audio
+        m.append("div").classed("divider", true);
+        m.append("div").classed("menu-section-label", true).text("Audio Server");
+        menuElement = m.append("div").classed("menu-element", true);
+        menuElement.append("div").classed("menu-label", true).text("Adress:")
+
+        menuElement
+            .append("input")
```
thequilo, line 645: Add `.classed("menu-control", true)` to get the correct alignment with the other menu entries

meeteval/viz/visualize.js
```diff
@@ -1624,10 +1642,37 @@ class CanvasPlot {
...
+                        trials.push(url.href);
+                    } else {
+                        trials.push(settings.audio_server + value);  // usually fails, but audio_server could define a prefix
+                        // trials.push(settings.audio_server + url.pathname);  // could this work?
+                        trials.push(url.href);
+                    }
+                }
+
+                console.log("Audio candidates", trials)
+
```
thequilo, line 1671: This code should be moved to a function and also used in `drawRecordingAudioButton`

meeteval/viz/visualize.js
```diff
@@ -1636,28 +1681,37 @@ class CanvasPlot {
                 let tooltip = addTooltip(element, false).classed("wrap-60 alignleft", true)
 
                 tooltip.append("div").attr('align', 'left').text(value)
+                let warning_field = tooltip.append("div").attr('align', 'left').style("display", "none").text(
+                    "\nIssue while loading\n  " + settings.audio_server + value +
+                    "\nWith access to the audio file (either via audio server from the menu, or directly from the filesystem), a player will appear.\n" +
+                    "Options:\n" +
+                    " - With 'python -m meeteval.viz.file_server' you can start a process, that exposes normalized wav files on http://localhost:7777\n" +
```
thequilo, line 1688: There is a `<code>` tag that formats code correctly (https://www.w3schools.com/tags/tag_code.asp)

meeteval/viz/visualize.js
boeddeker: drawRecordingAudioButton is different, it requires the audio_server. I don't know if and how to partially load audio.

Additionally, drawRecordingAudioButton is more verbose, so it feels more natural to try only one combination.

meeteval/viz/visualize.js
boeddeker: Done

meeteval/viz/visualize.js
boeddeker: Done.

meeteval/viz/visualize.js
thequilo: Hmm, okay. As a user, I would expect that both work the same. But you are right, as far as we know, partial loading is not possible without a server

thequilo (approved):

PR #76, opened 2024-05-15
Rework overview table

thequilo:
Reorders and groups columns from the same system. Numbers are now displayed with a monospaced font and aligned at the decimal dot. The sorting state is stored in the URL so that hitting the back button in the browser returns to the earlier state.

This is an example:
image
with the sort encoded like this in the URL: `.../index.html?sort=1&order=descending`


boeddeker:
What do you think about a table-striped style?

boeddeker (approved):

PR #77, opened 2024-06-10
Add testcase that runs code blocks in readme

thequilo:
Extracts and runs code blocks in README.md.

Tests the output of print statements against the following comment lines, ignoring any whitespace.

Only executes python cells and parses STM cells. All other cells are not tested.


thequilo:
We could also test bash/shell cells, but then we have to introduce a mechanism to exclude cells, e.g. `git clone` (don't we?)

boeddeker (approved):
Fine with the offline requested changes

PR #83, opened 2024-08-20
Add greedy orc WER

thequilo:
This PR adds the core greedy algorithm and the wrappers for ORC WER. I will add other pull requests for the time-constrained greedy algorithm and the DI-cpWER when this is merged.

I named it "greedy combination matching", similar to the "optimal combination matching" acronym from the ORC WER. Given this naming, I'm unsure if the name "greedy ORC-WER" fits well because it mixes the terms "greedy" and "optimal". But, at the same time, the term ORC-WER is already known, and greedy ORC seems clearer to me.

I did not use "reference" and "hypothesis" but "segments" and "streams" so that the core code can be used for both ORC-WER and DI-cpWER without name conflicts.

### DIFF
--- a/meeteval/wer/__main__.py
+++ b/meeteval/wer/__main__.py
@@ -189,7 +189,6 @@ def greedy_orcwer(
         regex=None,
         reference_sort='segment_if_available',
         hypothesis_sort='segment_if_available',
-        assignment_initialization='cp',
         uem=None,
         partial=False,
         normalizer=None,
@@ -203,7 +202,6 @@ def greedy_orcwer(
         uem=uem,
         partial=partial,
         normalizer=normalizer,
-        assignment_initialization=assignment_initialization,
     )
     _save_results(results, hypothesis, per_reco_out, average_out)
 
@@ -550,17 +548,6 @@ class CLI:
                      '- None: Do nothing (default)\n'
                      '- lower,rm(.?!,): Lowercase the transcript and remove punctuations (.,?!).',
                 choices=[None, 'lower,rm(.?!,)'],
-            )
-        elif name == 'assignment_initialization':
-            command_parser.add_argument(
-                '--assignment-initialization',
-                choices=['cp', 'random', 'constant'],
-                help='Specifies how the greedy search should be initialized.\n'
-                     'Choices:\n'
-                     '- cp: Use the assignment from the cpWER. Ensures fast convergence when the speaker labels '
-                     'are already somewhat correct. Requires speaker/stream labels in both, reference and hypothesis.\n'
-                     '- random: Initializes the assignment randomly. This is not deterministic.\n'
-                     '- constant: Initializes the label for all segments to 0.'
             )
         elif name == 'partial':
             command_parser.add_argument(
--- a/meeteval/wer/__main__.py
+++ b/meeteval/wer/__main__.py
@@ -189,7 +189,6 @@ def greedy_orcwer(
         regex=None,
         reference_sort='segment_if_available',
         hypothesis_sort='segment_if_available',
-        distancetype='21',
         assignment_initialization='cp',
         uem=None,
         partial=False,
@@ -204,7 +203,6 @@ def greedy_orcwer(
         uem=uem,
         partial=partial,
         normalizer=normalizer,
-        distancetype=distancetype,
         assignment_initialization=assignment_initialization,
     )
     _save_results(results, hypothesis, per_reco_out, average_out)
@@ -552,17 +550,6 @@ class CLI:
                      '- None: Do nothing (default)\n'
                      '- lower,rm(.?!,): Lowercase the transcript and remove punctuations (.,?!).',
                 choices=[None, 'lower,rm(.?!,)'],
-            )
-        elif name == 'distancetype':
-            command_parser.add_argument(
-                '--distancetype',
-                choices=['1', '2', '21'],
-                help='Specifies the costs used for substitutions in the greedy distance calculation.\n'
-                     'Choices:\n'
-                     '- 1: A substitution cost of 1 (standard Levenshtein distance)\n'
-                     '- 2: A substitution cost of 2 (subst = ins + del)\n'
-                     '- 21: Optimize with 2 until convergence, then optimize with 1 until convergence. This is the '
-                     'default and recommended for best results.'
             )
         elif name == 'assignment_initialization':
             command_parser.add_argument(
--- a/meeteval/wer/__main__.py
+++ b/meeteval/wer/__main__.py
@@ -182,6 +182,34 @@ def orcwer(
     _save_results(results, hypothesis, per_reco_out, average_out)
 
 
+def greedy_orcwer(
+        reference, hypothesis,
+        average_out='{parent}/{stem}_orcwer.json',
+        per_reco_out='{parent}/{stem}_orcwer_per_reco.json',
+        regex=None,
+        reference_sort='segment_if_available',
+        hypothesis_sort='segment_if_available',
+        distancetype='21',
+        assignment_initialization='cp',
+        uem=None,
+        partial=False,
+        normalizer=None,
+):
+    """Computes the Optimal Reference Combination Word Error Rate (ORC WER)
+    using a greedy algorithm. This algorithm is a lot faster than the optimal
+    algorithm, but IS NOT GUARANTEED TO PRODUCE THE SAME RESULT!"""
+    results = meeteval.wer.api.greedy_orcwer(
+        reference, hypothesis, regex=regex,
+        reference_sort=reference_sort, hypothesis_sort=hypothesis_sort,
+        uem=uem,
+        partial=partial,
+        normalizer=normalizer,
+        distancetype=distancetype,
+        assignment_initialization=assignment_initialization,
+    )
+    _save_results(results, hypothesis, per_reco_out, average_out)
+
+
 def cpwer(
         reference, hypothesis,
         average_out='{parent}/{stem}_cpwer.json',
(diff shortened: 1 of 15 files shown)


meeteval/wer/__main__.py
```diff
@@ -525,6 +553,28 @@ def add_argument(self, command_parser, name, p):
                      '- lower,rm(.?!,): Lowercase the transcript and remove punctuations (.,?!).',
                 choices=[None, 'lower,rm(.?!,)'],
             )
+        elif name == 'distancetype':
```
boeddeker, line 556: Should we really allow to change the value?
 - 1: It fails too often.
 - 2: It is unlikely to yield a desired alignment
 - 21: The only one, that we properly tested

meeteval/wer/__main__.py
```diff
@@ -525,6 +553,28 @@ def add_argument(self, command_parser, name, p):
...
+                '--distancetype',
+                choices=['1', '2', '21'],
+                help='Specifies the costs used for substitutions in the greedy distance calculation.\n'
+                     'Choices:\n'
+                     '- 1: A substitution cost of 1 (standard Levenshtein distance)\n'
+                     '- 2: A substitution cost of 2 (subst = ins + del)\n'
+                     '- 21: Optimize with 2 until convergence, then optimize with 1 until convergence. This is the '
+                     'default and recommended for best results.'
+            )
+        elif name == 'assignment_initialization':
```
boeddeker, line 567: Have you tried the alternatives? Is there a motivation to provide them?

random: We should not allow an option that yields non deterministic results

meeteval/wer/matching/greedy_combination_matching.py
```diff
@@ -0,0 +1,216 @@
...
+            substitution costs or with or without time constraint.
+    """
+    # Temporary variables
+    cost_differences = np.zeros((len(streams),), dtype=int)
+    costs_with_utterance = np.zeros((len(streams),), dtype=int)
+    costs_without_utterance = np.zeros((len(streams),), dtype=int)
+    updated_forward_columns = [None] * len(streams)
+
+    # Iteratively update assignment greedy
+    for num_updates in range(1, 50):
```
boeddeker, line 49: Add a comment, that 50 is chosen to be far too large?
If I remember correctly, num_updates didn't go larger than 3 or 4.

meeteval/wer/matching/greedy_combination_matching.py
```diff
@@ -0,0 +1,216 @@
...
+    if initialization == 'cp':
+        # Compute cpWER to get a good starting point
+        from meeteval.wer.wer.cp import _minimum_permutation_assignment
+        from meeteval.wer.wer.siso import siso_levenshtein_distance
+        assignment, _ = _minimum_permutation_assignment(
+            segments.groupby('speaker'),
+            streams,
+            distance_fn=siso_levenshtein_distance,
+        )
+        # Use integers for th assignment labels.
```
boeddeker, line 167: th -> the

meeteval/wer/matching/greedy_combination_matching.py
```diff
@@ -0,0 +1,216 @@
...
+        segments: A SegLST object containing the segments that are assigned to
+            the streams
+        streams: A SegLST object containing the streams.
+        initialization: How the initial stream labels are obtained before the
+            greedy algorithm starts. Can be one of:
+            - `'random'`: Assigns random initial stream labels to each segment
+            - `'constant'`: Assigns a constant stream label of 0 to each segment
+            - `'cp'`: Uses the cpWER solver to obtain the initial stream labels. Assumes
+                that the `'speaker'` key is present in both `a` and `b`.
+    """
```
boeddeker, line 157: Add some small doctests?

meeteval/wer/wer/orc.py
```diff
@@ -299,3 +309,114 @@ def apply_orc_assignment(
...
+
+    def matching(reference, hypothesis):
+        """Use the mimo matching algorithm. Convert inputs and outputs between the formats"""
+        distance, assignment = greedy_combination_matching(
+            reference.T['words'],
+            [[w for words in stream.T['words'] for w in words] for stream in hypothesis.values()],
+            initial_assignment=initialize_assignment(reference, hypothesis, initialization=assignment_initialization),
+            distancetype=distancetype,
+        )
+        if not str(distance).endswith('1'):
```
boeddeker, line 362: Should this be `if not distancetype.endswith('1')`?

boeddeker: Can you improve the comment? It took me some time to understand, what it means.

tests/test_greedy_combination_matching.py
```diff
@@ -0,0 +1,45 @@
...
+    """Burn-test. Brute-force is exponential in the number of reference
+    utterances, so choose a small number."""
+    distance, assignment = greedy_combination_matching.greedy_combination_matching(
+        segments, streams, [0] * len(segments)
+    )
+    _check_output(distance, assignment, segments, streams)
+
+
+@given(segments(6), streams(4))
+def test_greedy_bound_by_optimal(segments, streams):
```
boeddeker, line 38: Can you add some tests, where greedy is tested to be equal to the optimal algorithm?

meeteval/wer/__main__.py
thequilo: Agreed, I'll remove the argument

meeteval/wer/__main__.py
thequilo: I haven't tried them on large dataset, but even the constant initialization yielded good results for the examples I looked at. The motivation is that the ORC-WER does not need speaker labels, but the cp initialization needs them.

meeteval/wer/matching/greedy_combination_matching.py
```diff
@@ -0,0 +1,216 @@
...
+from meeteval.io.seglst import SegLST
+from meeteval.wer.matching.cy_greedy_combination_matching import cy_forward_col
+
+
+def _apply_assignment(assignment, segments, n=None):
+    if n is None:
+        n = max(assignment) + 1
+    else:
+        n = max(max(assignment) + 1, n)
+        assert n > max(assignment), (n, max(assignment))
```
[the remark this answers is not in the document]
thequilo: Done

meeteval/wer/matching/greedy_combination_matching.py
thequilo: Done

meeteval/wer/matching/greedy_combination_matching.py
thequilo: Done

tests/test_greedy_combination_matching.py
thequilo: Yes, I added it for the example files

meeteval/wer/__main__.py
boeddeker: An unbalanced initial assignment could have side effects.
But maybe the side effect is only, that the code takes some more iterations.

Has ORC-WER issues with cp? I'd like to have always the same init and in general it is best,
when nobody changes arguments to increase the probability, that people can resonably compare across publications (i.e. we don't want to have the asclite issue, where everyone has a slightly different evaluation).

tests/test_greedy_combination_matching.py
```diff
@@ -0,0 +1,79 @@
...
+@given(segments(10), streams(4), st.sampled_from(['1', '21']))
+def test_optimal_assignment_is_not_changed(segments, streams, distancetype):
+    """
+    Test that the optimal assignment is not changed.
+
+    Only works with distancetype='1'. The optimal ORC assignment is not
+    guaranteed to be optimal for other substitution costs than 1.
+    For distanctype='21' we can at least guarantee that the greedily corrected
+    assignment is not worse than the optimal assignment, but we cannot guarantee
+    equality.
```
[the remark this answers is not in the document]
thequilo: I just found an example where distancetype='21' changes the distance to be worse than the initialization. So, the argumentation is not correct.

Example:

```py
segments=[[2], [0, 1, 1, 1]]
streams=[[0, 0, 0, 0], [0, 1]]

# ORC assignment
assignment = [1, 0]
distance = 5

# After distancetype=2
assignment = [1, 1]
distance = 7

# After distancetype=1
assignment = [0, 1]
distance = 6  # Optimal distance of 5 is not recovered!
```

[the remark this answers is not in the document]
boeddeker: So you plan to disable '21' for this function completely and not just partially?
I tried to guess, based on the test code and your comment.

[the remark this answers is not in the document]
boeddeker: Hmm. Is it maybe possible to improve the greedy algorithm?
Wasn't there a paper, where `s, i, d = 4, 3, 3` was used to find an assignment?
Maybe it is worth testing that, e.g. `211 -> 433 -> 111` or `433 -> 111` instead of `211 -> 111`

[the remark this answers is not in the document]
thequilo: Tried that, `211 -> 433 -> 111` doesn't solve the problem for this example, `111` and `433 -> 111` solve this example correctly, but `433` does not help in all cases where `211` helps

meeteval/wer/__main__.py
thequilo: I removed the option. Now, cp is always used.

tests/test_greedy_combination_matching.py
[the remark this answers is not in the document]
thequilo: I'll merge this PR so that we can go forward with the DI-cpWER and greedy time-constrained variants

PR #94, opened 2024-09-17
add more normalizers, "lower,rm([^a-z0-9 ])" and chime6/7/8

boeddeker:
I changed the if statement in the source code to a class to have all code together that belongs to the normalizer.

The chimeX normalizers have a runtime dependency of chime-utils

### DIFF
--- a/meeteval/wer/api.py
+++ b/meeteval/wer/api.py
@@ -81,7 +81,7 @@ class _Normalizers:
         >>> test('chime7')
         'hello world äèア😊 hmmm hmmm hmmm wi-fi word-word 11 11 mr then said'
         >>> test('chime6')
-        'hello world äèア😊 hmmm hmmm hmmm wi-fi word-word 11 11 mr then said'
+        'hello world äèア😊 hm hmm hmmm wi-fi word-word 11 11 mr then said'
         """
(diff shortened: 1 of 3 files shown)


meeteval/wer/api.py
```diff
@@ -39,6 +39,87 @@ def _maybe_load(paths, file_format) -> meeteval.io.SegLST:
...
+            #     present.
+
+            def normalizer(seg):
+                words = chime_utils_normalizer(seg['words'])
+                if words != chime_utils_normalizer(seg['words']):
+                    raise RuntimeError(
+                        f'You discovered an idempotence bug in the chime_utils normalizer: {normalizer!r}.\n'
+                        f'Original:         {seg["words"]}\n'
+                        f'Normalized:       {words}\n'
+                        f'Normalized again: {chime_utils_normalizer(words)}'
```
thequilo, line 111: Add instructions for what the user should do if this happens?

thequilo (approved):

meeteval/wer/api.py
boeddeker: After an offline discussion, we decided to not include a link to the github chime_utils or meeteval issues.

PR #102, opened 2025-03-20
Add time-constrained MIMO-WER

thequilo:


meeteval/wer/matching/time_constrained_mimo_matching.h
```diff
@@ -0,0 +1,853 @@
...
+/*
+* A hash function for vectors so that they can be used as keys in unordered_map.
+*
+* This might change in the future. We might switch to a plain integer representation for indexes, similar to the
+* (not-time-constrained) mimo algorithm.
+*
+* https://stackoverflow.com/a/53283994
+*/
+struct VectorHasher {
+    int operator()(const std::vector<unsigned int> &V) const {
```
boeddeker, line 21: Hashs are used to quickly evaluate, if two things are different, but they usually aren't sufficient to check for equality.
Is that a problem?

boeddeker (approved):

meeteval/wer/matching/time_constrained_mimo_matching.h
thequilo: This is not a problem, equality is checked separately. From the docs (https://en.cppreference.com/w/cpp/container/unordered_map):

> Two keys are considered equivalent if the map's key equality predicate returns true when passed those keys. If two keys are equivalent, the hash function must return the same value for both keys.

PR #106, opened 2025-04-04
Add meeteval-wer normalize CLI command

thequilo:
Adds a command `meeteval-wer normalize` that normalizes the words in any input file according to one normalizer and dumps it in the same or any other format.

```console
$ meeteval-wer normalize example_files/ref.stm --normalizer 'chime8' -o -
recordingA 1 Alice 1 1 effort made was a lot i know you are busy we need to make the new version clean
recordingA 1 Bob 1 2 strategic staircase we need a paradigm shift
recordingA 1 Bob 2 3
recordingA 1 Alice 2 3 pipeline bob called an all hands this afternoon
recordingB 1 Alice 3 4 hop on the bandwagon let us schedule a standup during the sprint to review our kpis
recordingB 1 Alice 4 5 hop on the bandwagon let us schedule a standup during the sprint to review our kpis
recordingA 1 Alice 5 6 organic growth what do you feel you would bring to the table if you were hired for this position it just needs more cowbell yet groom the backlog land the plane
```

The default is set to `chime8`. I'm not sure if this is the best choice for the default. I may make the `--normalizer` option mandatory.

### DIFF
--- a/meeteval/io/smart.py
+++ b/meeteval/io/smart.py
@@ -177,17 +177,30 @@ def dump(obj, path, format: 'str | None'=None, force=True):
     
     format_guessed = format in (None, 'none', 'auto')
     if format_guessed:
-        format = _guess_format(Path(path))
+        if str(path) == '-':
+            format = None
+        else:
+            format = _guess_format(Path(path))
 
-    dumper = _get_format(format, path)
+    if format is None:
+        import meeteval
+        dumper = obj.__class__
+        assert issubclass(dumper, meeteval.io.base.BaseABC), dumper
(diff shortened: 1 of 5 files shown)


boeddeker:
Hmm. I am not sure, whether chime8 is a reasonable default. It does a lot, and it is not clear yet, whether it will be adopted beyond chime8.

I have a preference to use a default normalizer that doesn't change an externally normalized reference file (except maybe upper/lower). So for example the normalizer that uses lower and removes punctuation would only in very rare cases change the reference.
The only case, that came to my mind is, when no normalizer should be used.

An alternative would be to use no default, as you mentioned.

PR #107, opened 2025-04-11
refactor: using h_minus_r_set to lookups

allrob23:
While checking the `apply_multi_file` function in `meeteval/io/seglst.py`, I noticed `h_minus_r` was a list used for checking if keys are in it during a dictionary comprehension. List lookups are O(N), so I refactored it to use a set for faster O(1) lookups.

Now, `h_minus_r_set` holds the set of keys in `hypothesis` not in `reference`, and the comprehension uses it to filter keys. I kept `h_minus_r` as a list for logging (like showing `h_minus_r[:5]`), so all output and behavior stay exactly the same. It's not a huge optimization, but it makes key checks a bit more efficient, especially if `h_minus_r` has many keys.

### DIFF
--- a/meeteval/io/seglst.py
+++ b/meeteval/io/seglst.py
@@ -625,7 +625,8 @@ def apply_multi_file(
     # Check session keys. Print a warning if they differ and raise an exception
     # when they differ too much
     if reference.keys() != hypothesis.keys():
-        h_minus_r = list(set(hypothesis.keys()) - set(reference.keys()))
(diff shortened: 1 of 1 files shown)


thequilo:
Hi @allrob23,

thank you for bringing this up! Now that I looked again at the code, I realized that the dict comprehension is not needed at all because further below, only the entries from the hypothesis dict are accessed that are in the reference keys.

So, you can remove the dict comprehension entirely.

allrob23:
Hello @thequilo, I committed now, is it ok?

PR #111, opened 2025-05-09
Add SISO WER to high-level Python api

### LINKED ISSUE #110: Is there a high-level python interface for the standard WER?
tcpwer has a high-level python interface which can be used as follows:
`wers = meeteval.wer.tcpwer('example_files/ref.stm', 'example_files/hyp.stm', collar=5)`
Is there a high-level python interface for the standard WER? 
I want to use the reference file and the hypothesis file as inputs to compute the standard WER by a high-level python interface.
But I found no interface for the standard WER in `meeteval/wer/api.py`.
The low-level interface for the standard WER is not able to take files as input.
`wer = meeteval.wer.wer.siso.siso_word_error_rate(reference='XXX', hypothesis='XXX')`
So, how can I use the high-level interface of the standard WER in the same way as using the high-level interface of tcpwer?

thequilo:
Closes #110.

Add the SISO WER as `meeteval.wer.sisower` to the high-level Python api.

I named it `sisower` instead of `wer` to avoid conflicts with the module `meeteval.wer.wer`. I didn't change the CLI command name, it is still just `wer`.

The new API function now also supports STM and SegLST files when the file names are unique. Previously, only keyed text files were supported. This was done to avoid accidentally calling the SISO WER when you wanted to compute, e.g., the cpwer. But I now think that it is very unlikely to have unique filenames / session_ids when you don't want to compute the SISO WER.

### DIFF
--- a/example_files/hyp_siso.stm
+++ b/example_files/hyp_siso.stm
@@ -1,0 +1,4 @@
+recordingA 1 Alice 0 1 Effort made was a lot i know your busy we need two make the new version clean
+recordingA 1 Bob 1 2.5 Strategic staircase we need a paradigm shift
+recordingC 1 Bob 2.1 3 Pipeline Bob called an all-hands this afternoon
+
--- a/example_files/ref_siso.stm
+++ b/example_files/ref_siso.stm
@@ -1,0 +1,3 @@
+recordingA 1 Alice 1 1 Effort made was a lot i know you're busy we need to make the new version clean
+recordingA 1 Bob 1 2 Strategic staircase we need a paradigm shift
+recordingC 1 Bob 2.1 3 Pipeline Bob called an all-hands this afternoon
--- a/meeteval/io/seglst.py
+++ b/meeteval/io/seglst.py
@@ -221,7 +221,7 @@ class SegLST(BaseABC):
         """
         Returns the unique values for `key` among all segments.
         """
-        return set([s[key] for s in self.segments])
+        return set(self.T[key])
 
     def __iter__(self):
         return iter(self.segments)
@@ -567,7 +567,8 @@ def apply_multi_file(
         reference, hypothesis,
         *,
         allowed_empty_examples_ratio=0.1,
-        partial=False
+        partial=False,
+        _groupby=None,
 ):
     """
     Applies a function individually to all sessions / files.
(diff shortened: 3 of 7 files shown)


boeddeker:
Does the code work for unique (filename speaker) combinations, i.e., I'd like to avoid the speech separation issue from kaldi?
Does the code fail for invalid inputs?

thequilo:
The code currently only looks at filenames (session_id). It would be possible to add an argument to also look at speakers (or any other key), but I wouldn't make it the default.

The code fails when there is more than one entry for any session_id and if the session_ids from reference and hypothesis do not match exactly.

thequilo:
For the issue with unique (filename speaker) combinations, I implemented such an option (naming is not ideal):

`meeteval-wer wer -r ref_siso.stm -h ref_siso.stm --keys session_id speaker`

Which then treats everything with a unique (session_id speaker) combination as one item. If you only want session_id, you can write `--keys session_id` and you can even add more keys and custom keys (with SegLST) like `--keys session_id speaker channel`.

What do you think about this?

boeddeker:
Hmm. So you mean, the code should not fail, when there is a placeholder in hyp for the speaker ID?
I am a bit biased from kaldi. From what I remember, every filename has the true speaker identity prefixed.
So the information is available in those cases.
Nonetheless, it would be annoying, if the code fails, when the speaker id is missing.

How about some "magic", because I want to avoid a command line argument, where you cannot select something. It just has to be the correct value.

If every session_id/filename in the reference has one segment use the session_id. 
 - The code will work, when the user has set the speaker_id to be the true speaker (this should never fail, independent of what we do)
 - The code will work, when the speaker_id is a placeholder or some random value
 - The code will work, when the speaker_id is not present

If the criterion above failed, use `(session_id, speaker)` as identifier:
 - The code will work, when the user uses the speaker key properly
 - The code will complain, when the speaker_id is a placeholder or missing.
         - In this case, it is likely, that the file is bugged
 - The code will complain, when the speaker_id is an estimated speaker id (spk1 vs Alice)
        - There is a permutation problem between ref and hyp. That will not be solved by siso. The user could use cpwer in those cases. Although the concatenation is not used.


Motivation to support speaker id by default:
 - If we implement the discussed log folder, that folder will have input files with the `(session_id, speaker)` identifier. So checking the WER would then be just a sisower call.
 - Source separation or recognition system for mult-talker files can be directly evaluated.
 - I don't know a case, where an exception would help to detect an error with this logic or a case, where a bug is hidden.

boeddeker (approved):
Could you add a burn test?

PR #112, opened 2025-06-12
Fix that the alignment calculation for the visualization didn't use point estimates

boeddeker:
Fix the mismatch between shown WER and alignment for corner cases of time constraint visualization.

The shown WER uses the point estimates for hyp (same as cli default), but the alignment calculation didn't use the point estimates.

This has only an effect for time constraint WERs and is only noticeable, when the time unit in the input files is not second,
e.g., a collar of 5 samples is far to low, while a collar of 5 seconds is a good choice, where the issue is barely noticeable.

How can this be observed:
image
In the tooltip to the WER, the count column is correct for the WER, but the speaker statistics reflect the alignment.
When there is a mismatch, it indicates, that some correct/substitution assignments are at the edge of the collar.

Should we add an assert to the code, that the statistics match? I don't know, if the same sub/ins/del is reported, since they are not unique.

@tobvogel Thanks for the report.

### DIFF
--- a/meeteval/viz/visualize.py
+++ b/meeteval/viz/visualize.py
@@ -436,7 +436,7 @@ def get_visualization_data(ref: SegLST, hyp: SegLST, assignment='tcp', alignment
 
     data['info']['wer_by_speakers'] = {
         speaker: wer_by_speaker(speaker)
-        for speaker in list(ref.unique('speaker'))
+        for speaker in list((ref + hyp).unique('speaker'))
     }
     for k in [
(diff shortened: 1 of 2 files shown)


thequilo:
Thanks for addressing this. Your solution is fine as long as the alignment stays hard-coded. We have to use a different approach when we make it configurable.

The align function should produce the same statistics as the wer computation. The core function is tested against kaldialign, only the preprocessing may make a difference.

I'm in general in favor of adding an assertion that the wer statistics equal the alignment statistics

thequilo (approved):

PR #116, opened 2025-06-24
Added pseudo alignment strategy based on phoneme duration 

popcornell:
Hi guys, 

Greetings from Brno. 
I am trying to add phoneme-based duration as an another pseudo-alignment word duration strategy. 
This could enable tcpWER for languages such as Japanese for which one character e.g. a kanji could be of much longer duration than others. 
Adding here also Alexander Polok as he is responsible for https://huggingface.co/spaces/BUT-FIT/EMMA_leaderboard
@Lakoc

### DIFF
--- a/meeteval/wer/wer/time_constrained.py
+++ b/meeteval/wer/wer/time_constrained.py
@@ -5,6 +5,8 @@ import itertools
 import typing
 from dataclasses import dataclass, replace
 
+import transphone
+
 from meeteval.io.pbjson import zip_strict
 from meeteval.io.stm import STM
 from meeteval.io.seglst import SegLST, seglst_map, asseglst, SegLstSegment
@@ -40,7 +42,7 @@ __all__ = [
 
 
 # pseudo-timestamp strategies
-def equidistant_intervals(interval, words):
+def equidistant_intervals(interval, words, *args):
     """Divides the interval into `count` equally sized intervals
     """
     count = len(words)
@@ -57,7 +59,7 @@ def equidistant_intervals(interval, words):
     ]
 
 
-def equidistant_points(interval, words):
+def equidistant_points(interval, words, *args):
     """Places `count` points (intervals of size zero) in `interval` with equal distance"""
     count = len(words)
     if count == 0:
@@ -72,7 +74,7 @@ def equidistant_points(interval, words):
     ]
 
 
-def character_based(interval, words):
+def character_based(interval, words, *args):
     """Divides the interval into one interval per word where the size of the interval is
     proportional to the word length in characters."""
     if len(words) == 0:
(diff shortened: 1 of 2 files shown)


meeteval/wer/wer/time_constrained.py
```diff
@@ -5,6 +5,8 @@
 import typing
 from dataclasses import dataclass, replace
 
+import transphone
```
boeddeker, line 8: Can you use a lazy import? We want to keep the mandatory dependencies for the core code small.

thequilo:
Hey @popcornell!

This is a good extension for evaluating such languages as Japanese. 

I'm unsure about your choice for the interface. I'm not that happy to add a `language` argument to that many functions just for this one pseudo word level timestamp strategy. This would have to be added to the whole interface including `api` and the CLI. I prefer something like `strategy='phoneme_based_jpn'` or `strategy=phoneme_based('jpn')`, so that the interface doesn't change.

meeteval/wer/wer/time_constrained.py
```diff
@@ -93,7 +95,30 @@ def character_based(interval, words):
     ]
 
 
-def character_based_points(interval, words):
+def phoneme_based(interval, words, language):
+    """Divides the interval into one interval per word where the size of the interval is
+    proportional to the number of phonemes in the word."""
+
+    g2p = transphone.read_tokenizer(language)
```
boeddeker, line 102: The call transphone.read_tokenizer sounds, that it loads a model. Has it a caching? 
If not, we should do a caching, since this function is called for every segment.

meeteval/wer/wer/time_constrained.py
```diff
@@ -40,7 +42,7 @@ class Segment(TypedDict):
 
 
 # pseudo-timestamp strategies
-def equidistant_intervals(interval, words):
+def equidistant_intervals(interval, words, *args):
```
boeddeker, line 45: I would prefer `(interval, words, language)` as signature, or at least `(interval, words, **kwargs)`.

popcornell:
> For Japanese often the character error rate is reported, but I don't know if usually the tools are language aware or people prepare the transcript to use WER calculators. Depending on your answer, the language argument could be used at multiple positions.

yeah actually I was unsure about that too. I was assuming that one would split the reference before feeding to meeteval but yeah maybe the best way to handle this is to make it dependent on the language or have an additional argument.
I am not sure if we should handle it depending on language though because the reference and/or system might have the whitespaces or may not.

thequilo:
@popcornell Do you have examples for the output of a Japanese ASR system? The guys from NTT said that CER is usually used instead of WER, which completely ignores whitespace and splits individual characters. In that case, we may want to add a time-constrained CER

thequilo:
We discussed the following:

**The language should be encoded in the strategy name.**
We found no other use case for a language argument than this alignment strategy. The distinction between CER and WER should be explicit and independent of the language. So, we want the language to be encoded in the alignment strategy key, like `transphone_phoneme_based_jpn`. For this, the `pseudo_word_level_strategies` should become a class so that it can split off the language and pass it on to the transphone library. Simlar to the normalizer in https://github.com/fgnt/meeteval/blob/main/meeteval/wer/normalizer.py.

Since there are potentially multiple libraries for obtaining phoneme durations, the package name should also be encoded in the strategy name.

**CER in a different PR**
We'll do the split into characters in a different pull request. In that PR, we'll for now supply a short script that splits segments into words as a preprocessing so that the WER functions yield a CER.

@popcornell Are you willing to adjust the PR with the required changes or should we do it?

popcornell:
Hey guys, yeah I plan to adjust it. But I am currently busy in the JSALT I thought I would have more time. I can do it this weekend though.

